Issue #699, opened 2026-08-11
flow:register has no wave-policy template: implementation authority, gate policy and file lanes live in retyped prose that does not survive a worker's /clear


cooneycw:
## Summary

`/flow:register` records a role, an address, and a lane (cwd/repo/issue/branch). Everything else a wave runs on — implementation authority, gate policy, who judges, the ledger format, the file lane, the authority model — lives in prose the orchestrator retypes into every worker message. That prose is where this wave's orchestration errors happened, and it does not survive a worker's `/clear`, which is the one thing the registry was built to do.

Proposal: a registration template with two tiers — **wave-level policy** declared once by the orchestrator and INHERITED by workers, and **role-level facts** per session — answerable non-interactively.

## Evidence from the 2026-08-11 `cpp` wave

Every item below is a thing that actually happened while running an 11-issue wave across two workers.

**1. The orchestrator's worst error came from lane data living in prose.** A brief fenced CPP-2 out of `scripts/flow-helpers-install.sh` and then, three paragraphs later in the SAME message, assigned it #677 — the issue whose fix is in that file. The worker refused to start and asked which. The cause was not carelessness; it was that the file lane existed only as hand-retyped text, with nothing to read it back from.

**2. The wave name was INFERRED.** A worker was started with only a role label. It reasoned its way to the right wave by checking which wave had a live orchestrator and no workers — and said so, flagging the inference. It was correct. That is luck, and #671 exists precisely because guessing wrong is silent.

**3. Implementation authority was established entirely out-of-band.** The single most consequential wave-level fact — whether this wave writes code or files issues, given CPP's standing "file issues, don't implement" rule — required a user round-trip and then had to be restated in both worker briefs. Both workers independently held the standing rule and correctly refused to infer authority from being handed an issue number. Nothing in the registry could have told them.

**4. The authority model nearly deadlocked the wave.** A worker asked whether it needed BOTH its user's go and the orchestrator's before writing code, and held pending an answer. Requiring two signatures produces exactly the stall #676 documents. This is a wave-level policy that should be declared once, not renegotiated per worker.

**5. The brief was retyped into every message and does not survive compaction.** The registry's documented purpose is surviving a worker's `/clear`. The gate points, ledger format and pushback rule — the actual protocol — live outside it, so a compacted worker loses exactly the rules the registry was built to preserve.

[...]

Issue #697, opened 2026-08-11
Test convention: a negative-condition fixture should assert its own precondition


cooneycw:
## Summary

A test fixture that creates a NEGATIVE condition indirectly — by constructing an environment rather than removing one named thing — can succeed at something broader than it intended, making the test pass while proving nothing. Propose adopting the one-line guard that catches it: **a negative-condition fixture asserts its own precondition.**

This is preventive, not corrective. See the sweep below — there is currently exactly one instance in the suite and it is already fixed.

## The instance (found and fixed in #695)

Testing that `cpp-commands-link.sh --check` fails open when git is absent, the first fixture emptied `PATH`. That removes `git` — and also `ln`, `mkdir`, `readlink`, `bash`. The script then failed for reasons unrelated to git, and **the assertions still passed**: no advisory printed, exit code unchanged. Both are exactly what a correct fail-open produces.

A fail-open test that cannot distinguish "git absent" from "everything absent" is not testing fail-open. It was caught only because it went red for the wrong reason first; a green run would have been unfalsifiable from the outside, and invisible in review.

The shipped fixture builds `PATH` from explicit coreutils symlinks and then asserts the condition it claims to create:

```python
assert shutil.which("git", path=str(stub_path)) is None, "fixture must lack git"
```

That assertion is the actual defence. The symlink construction is just how this particular precondition gets built; the guard is what proves it was built correctly.

## Scope: narrower than it first appears

A sweep of `tests/` found no second instance, and the distinction is worth recording so the convention does not get over-applied:

- **`monkeypatch.delenv("X", raising=False)` is fine as-is.** It removes exactly one named variable, deterministically. Direct removal needs no guard.
- **`test_flow_finish_gate.py:71,282` are fine.** They *prepend* to `PATH` (`f"{bindir}:{env['PATH']}"`), which adds a stub without removing anything.
- The `assert not ... exists()` lines across `test_codex_skill_sync.py` and `test_cpp_commands_link.py` are **outcome** assertions about what the code under test produced, not fixture preconditions. Different thing.

So the risk is specific: **indirect** construction of an absence, where the mechanism is broader than the intent.

## Requested

Adopt as a test convention, documented wherever test conventions live:

> When a fixture creates a negative condition (a tool absent, a path missing, a capability unavailable) by constructing an environment rather than removing one named thing, assert the precondition holds before exercising the code under test. `assert shutil.which("git", path=...) is None, "fixture must lack git"` is the shape.

[...]

Issue #686, opened 2026-08-11
flow-helpers-install.sh --help prints past the doc block: sed range 2,37p over a header ending at line 28


cooneycw:
## Summary

`scripts/flow-helpers-install.sh --help` prints the doc block **plus** the first few lines of executable code. The extractor is:

```sh
sed -n '2,37p' "$0"
```

but the doc comment block ends at **line 28**. Lines 29-37 are code, so `--help` emits `set -uo pipefail`, three variable assignments, and two stray comment lines after the real help text.

## Reproduce

```sh
scripts/flow-helpers-install.sh --help | tail -8
```

Observed: the tail of the output is shell source, not documentation.

## Fix

One character range change: `sed -n '2,28p'`.

A more durable form, since the block will drift again the next time anyone adds an env var to the header, is to terminate on the first non-comment line rather than a hardcoded bound:

```sh
sed -n '2,${/^[^#]/q;p;}' "$0"
```

## Provenance

Found by the CPP-2 worker session while landing #677 (PR #684), and confirmed **pre-existing** rather than introduced: `git show origin/main:scripts/flow-helpers-install.sh` is byte-identical in the affected region, and PR #684's diff on that file is `+17/-0` — pure addition past line 37, so nothing above it moved.

Deliberately left unfixed in #684 under the wave rule that a new finding outside an issue's scope gets its own ruling rather than a drive-by fix. Filed here so it is not lost.

## Severity

Cosmetic. The help text is still present and correct; it just has code appended. Filing because `--help` is the discoverability surface for a script whose whole purpose is unattended/scripted installation, and because it is a one-character fix.

Issue #676, opened 2026-08-11
flow:wave has no orchestrator->worker delivery lane - formalize the wave-dir mailbox with monitor-based wake (human relay was the only transport)


cooneycw:
## Summary

In the 2026-08-11 aws-learn `cpp-install` wave, every orchestrator->worker delivery - the registration ack, the wave brief, the issue assignment - could only reach the worker when the USER typed a pointer into the worker's terminal by hand. The assignment sat fully-written but undelivered for ~2 hours while both sessions idled, each correctly "standing by." The wave protocol needs a first-class host-local delivery lane that does not assume symmetric session messaging, plus a wake mechanism, because a delivery lane an idle session never reads is not a lane.

## Observed constraints (all from one live wave, host learn27)

1. **Orchestrator->worker session messaging does not exist in the harness.** `SendMessage` from the orchestrator rejected all four identifier forms (observed `bridge:` address, raw session id, session UUID, display label) - it only routes to subagents the calling session spawned. Worker->orchestrator worked fine throughout (#672).
2. **Display labels rot.** A send addressed by label succeeded, then the same name failed minutes later after the label silently changed mid-session (#672 comment).
3. **Idle sessions cannot wake themselves.** The worker registered, sent its hello, and stood by - correctly. It re-reported "registered" on every user wake, but had no way to learn an assignment existed. Registration state, worktrees, branches all sat unchanged for 2h; the only missing element was delivery.
4. A workaround mailbox (`$XDG_RUNTIME_DIR/cc-flow-wave/<wave>/outbox-<role>.md` + an orchestrator-side watch on `inbox-*.md`) worked - but only after the user relayed "read your outbox" once per worker, because nothing in `/flow:register` tells a worker the mailbox exists or arms a watch on it.

## Requested: make the wave-dir mailbox a protocol lane, with monitor-based wake

Beside the #638 registry (same lifetime, same host-local scope):

- `outbox-<role>.md` - orchestrator writes ack/brief/assignment/verdicts; one file per role, rewrite-in-place with a rev marker.
- `inbox-<role>.md` (or `inbox-orchestrator.md`) - workers write hellos/reports when direct messaging fails.
- **Wake contract**: at `/flow:register` time the worker ARMS a persistent file watch on its own outbox (the harness Monitor tool, or a background `until`-loop on mtime); the orchestrator likewise watches `inbox-*.md`. A mailbox write then WAKES the counterpart - no human relay, no polling-on-hope.
[...]

PR #708, opened 2026-08-11
feat(flow): reserved lexicon for wave state transitions, validated at send and at record (Closes #701)

### LINKED ISSUE #701: flow:wave has no lexicon for state transitions: gate verdicts, lane grants and merge authorizations are free prose, and every wave miscommunication happened there
## Summary

The wave protocol has a strict lexicon at the machine-to-machine layer and free prose at the agent-to-agent layer. `FLOW_WAVE: registered|refused|verified|address_filled|mismatch-corrected` plus `FLOW_WAVE_*=` detail lines and exit codes never misfired across an 11-issue wave. Every miscommunication in that wave happened in the prose layer.

Proposal: a small reserved vocabulary for the speech acts that are **state transitions with a wrong-answer cost** — gate verdicts, lane grants, merge authorisations, state assertions — leaving the reasoning in prose where it belongs.

Companion to #699, which covers declared STATE. This covers TRANSITIONS.

## Evidence from the 2026-08-11 `cpp` wave

**1. A lane grant was read as a gate approval, and only caution prevented it landing.** One orchestrator message contained both "you are unblocked for Step 4" and "Hard stop at Step 3 stands for the gate report." It had been composed before the worker's gate report arrived. The worker held, and said why:

> Those two only reconcile one way: the LANE is unblocked, the GATE is not yet judged. Treating the lane approval as gate approval would be precisely the silent pass-through of Step 3 your brief names as the failure this gate exists to prevent.

It was right. Nothing but the worker's care stopped an unjudged gate from proceeding. A reserved `GATE:` / `LANE:` distinction makes that mechanical instead of interpretive.

**2. A lane fence was read as a lane grant.** A brief listed a file under "Explicitly NOT yours" for worker A, annotated with the issue that owns it — and the worker assigned that issue read the same file as belonging to A. Cost a full round-trip and a refusal to start. Off-limits-to-A and belongs-to-A are opposite facts expressed in adjacent prose.

**3. Wave-state broadcasts went stale four times.** Each was accurate when composed and wrong when read, because none carried an as-of stamp. Two were caught by workers ("#677 is already CLOSED, not 'landing now'"), one by the orchestrator, one corrected mid-thread. A mandatory `as-of <commit>` on every state assertion converts a silent staleness into a visible one.

**4. Conditional merge authorisation was invented ad hoc and repeated four times.** "Merge when `ci/woodpecker/pr/woodpecker` reports pass; on fail, STOP and report" was retyped in prose each time, with the precise check name that distinguishes it from the push pipeline. It is a recurring speech act with a machine-checkable predicate, and it was carried entirely by the orchestrator remembering to restate it.

**5. A merge was authorised outside the orchestrator chain and had no way to be recorded.** The repo owner typed "approved" directly into a worker session while it held for the orchestrator's go. The worker merged — correctly, the human outranks the protocol — and then flagged it precisely because there was no form for it:

> a merge that happened without your ruling is exactly the kind of thing that makes a wave-state broadcast wrong an hour later.

**6. A deviation-and-ratification had no form either.** A worker implemented an approved boundary, found it broke two pre-existing tests, shipped a narrower one and reported before committing. The orchestrator ratified. That exchange — deviate, justify, ratify or restore — is a distinct transition that occurred with no shared shape.

**7. Counter-evidence, and the strongest argument for the proposal.** The ONE structured element in the brief — the `delivered / in-scope / residual` ledger — produced directly comparable output from two workers across eleven issues, with residuals explicitly ruled rather than lost. The part that was structured did not drift. Nothing else was structured, and everything else did.

## Proposed scope

Reserved tokens for transitions only:

| Token | Carries |
|---|---|
| `GATE: GO` / `HOLD` / `GO-WITH-CONDITIONS` | the conditions, explicitly, when conditional |
| `LANE: GRANT` / `EXTEND` / `REVOKE` | the file lane — distinct from any gate verdict |
| `MERGE: AUTHORIZED` | the predicate, named (which check, not "CI") |
| `STATE: as-of <commit>` | mandatory on any wave-state assertion |
| `RATIFY` / `OVERRULE` | response to a reported deviation |
| `PUSHBACK` | so it cannot be skimmed past as agreement |
| `LEDGER` | already proven in practice; formalise the existing shape |

## What it must NOT cover

The reasoning. The highest-value messages in this wave were a worker's design-ruling requests, its correction of credit the orchestrator had misattributed to it, and its catch of the crossed lane/gate message. None would survive schematisation, and a lexicon that crowds them out costs more than it saves. Tokens carry the transition; prose carries the argument.

## The condition that decides whether this is worth building

**A lexicon nobody validates is prose with extra steps.** If `GATE: GO` becomes a token emitted reflexively while the real semantics stay in the paragraph below it, the result is a marker whose broken version is indistinguishable from its working version — the exact defect class this wave spent the day removing (#673, #674, #677, #685, #698).

Applying the wave's own rule to the protocol itself: *ask what a broken version of the measurement would print.* A reflexive `GATE: GO` prints what a considered one prints.

So this should ship only with something that reads it back. Minimally one of:

1. a registry verb that validates a message block against the vocabulary; or
2. a rule that a gate cannot be RECORDED as judged without a parseable verdict — making the token load-bearing rather than decorative.

Without one of those, prefer not building it.

## Relationship to other issues

- **#699** — declared state (wave policy, file lanes, authority). This is the transitions half of the same program.
- **#676** — if a mailbox delivery lane is built, it needs a message schema; this would be it. Argues for sequencing this before #676's redesign rather than after.

## Provenance

Proposed by the wave `cpp` orchestrator after an 11-issue two-worker wave on claude-power-pack, 2026-08-11. Every numbered item is an observed event with a quotable transcript, not a hypothetical.

cooneycw:
## Summary

The wave protocol is strict machine-to-machine and free prose agent-to-agent. `FLOW_WAVE: registered|refused|verified|address_filled|mismatch-corrected` never misfired across an 11-issue wave. **Every** miscommunication in that wave happened in the prose layer.

This adds a small reserved vocabulary for the speech acts that are **state transitions with a wrong-answer cost**, and leaves the reasoning in prose where it belongs.

Companion to #699, which covers declared STATE. This covers TRANSITIONS.

## The vocabulary

| Token | Enforced requirement | The failure it closes |
|---|---|---|
| `GATE: GO #N` | names its subject issue | a verdict whose subject is implied goes stale between composing and reading |
| `GATE: HOLD #N behind #M` | names what it waits behind | a hold with no blocker cannot be superseded on evidence |
| `GATE: GO-WITH-CONDITIONS #N` | `- <condition>` lines beneath it | conditions left in the paragraph below **is** the crossed lane/gate message |
| `LANE: GRANT\|EXTEND <role> <paths>` | role AND paths | a fence ("explicitly NOT yours") was read as its own inverse |
| `MERGE: AUTHORIZED #N when <check>` | a NAMED check | `ci/woodpecker/pr/woodpecker` is not the push pipeline; "when CI passes" is refused |
| `STATE: as-of <commit>` | the stamp is mandatory | four broadcasts were true when composed and wrong when read |
| `RATIFY \| OVERRULE #N <reason>` | issue + reason | a ruling with no reason is indistinguishable from inattention |
| `PUSHBACK <argument>` | must carry its argument | so it cannot be skimmed past as agreement |
| `LEDGER` | `delivered:`/`in-scope:`/`residual:` | the one structured element that did **not** drift, formalised rather than invented |

Reserved lines are LINE-ANCHORED (the #607 rule), so a mention inside a sentence never becomes a declaration - a design discussion of the lexicon still sends.

## The anti-decoration contract

The issue set its own kill condition: *a lexicon nobody validates is prose with extra steps - a reflexive `GATE: GO` prints what a considered one prints.* It asked for **one** read-back mechanism; this ships **both**, because either alone leaves a hole:

1. **`flow-wave-mailbox.sh send` validates every body.** A malformed transition cannot be DELIVERED - exit 6, `FLOW_MAILBOX: refused`, box untouched. Without this, only gate verdicts are covered and every lane grant stays prose.
2. **`flow-wave-lexicon.sh record` DERIVES the #645 ledger entry** from the parsed `GATE:` token (`ruling`, `holds_behind`, `serializes:` -> `adds_serialized`) instead of accepting a hand-written one. A gate cannot be recorded as judged without a parseable verdict. Without this, the token goes decorative the moment someone writes the ledger by hand.

**The token changes planner OUTPUT, not a log line.** `flow-wave-plan.py --verdicts` exits 4 on an unsuperseded hold and unions `adds_serialized` into `serialized_resources`. Proved by running the real planner twice over identical issue data:

```
control (no ledger):        exit 0,  serialized_resources {}
with a recorded HOLD:       exit 4,  verdict_conflicts [{issue: 52, ruling: hold}]
with a recorded serializes: exit 4,  serialized_resources {"migration-0009": [60, 61]}
```

That last one is the two-`0009`s migration collision - a condition changed #60's footprint though its body never did - which in the reference wave was caught only by a worker's status report.

## Two properties that keep this from becoming a stall

A validator that can block delivery is a new way for a wave to go quiet, which is precisely the #676 failure this lane exists to remove. So:

- **Absence is never an error.** A message with no reserved token is `none`/exit 0 and always delivers. Only a malformed PRESENT token refuses.
- **A missing or broken validator FAILS OPEN.** Only exit 1 ("a token is present and malformed") refuses; every other non-zero delivers with a note. Pinned by two tests.

`--no-lexicon` is the per-send escape when a prose line happens to open with a reserved word.

## What it deliberately does NOT cover

The reasoning. The highest-value messages in the reference wave were a worker's design-ruling requests, its correction of credit the orchestrator had misattributed to it, and its catch of the crossed lane/gate message. None would survive schematisation, and a vocabulary that crowds them out costs more than it saves. Tokens carry the transition; prose carries the argument.

## Test plan

- `~/.claude/scripts/flow-finish-gate.sh` -> `FLOW_FINISH_GATE: ok` (4/4 steps: lint, test **1622 passed / 0 failed**, mypy clean over 152 files, security clean)
- `make negative-fixture-check` -> ok (the gate #697 landed mid-implementation; this suite is exactly the shape it guards)
- `make codex-skills-check` -> 75 skills in sync
- 49 new tests. **The negatives carry the suite**, because a file that only feeds a parser well-formed messages would pass identically against a validator that accepted everything: 17 parametrized malformed shapes each asserted REFUSED with the refusal naming its line, plus positive controls (the same sentence passes once the check is named), negative controls (a plain approval carries neither optional key), and the load-bearing planner round-trips above.
- **Mutation-proved**: with `add_err` stubbed to a no-op - a validator that refuses nothing - **23 of 49 tests fail**. The suite detects the decorative-guard failure rather than assuming it away.

## Incidental fix

`flow-wave-mailbox.sh --help` printed a hand-counted `sed 2,90p` range, and adding the lexicon-gate paragraph pushed the header to 93 lines - #686 recurring the instant a header grew. Both this helper and the new one now stop at the first non-comment line, so the range cannot drift again.

## Residual (deliberately not in this PR)

`register.md`'s worker-side `PUSHBACK`/`LEDGER` guidance was dropped from the approved scope: #699 held `register.md` in a live session when this work was planned. #699 has since merged, so the collision is gone but the file was rewritten substantially - building on it unread would be the stale-base mistake in documentation form. Filed as a follow-up rather than left as a hole. Everything else in the plan shipped.

Closes #701

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


cooneycw:
Residual from the Step-3 scope call is now filed as #709 (worker-side `PUSHBACK`/`LEDGER` documentation in `register.md`).

Issue #769, opened 2026-09-05
flow-finish-gate: re-run only the failed test ids once before declaring fail


cooneycw:
## Problem

`flow-finish-gate.sh` declares `FLOW_FINISH_GATE: fail` on the first non-zero test exit, which is a hard STOP for `/flow:auto` Step 6. On a host running several concurrent sessions, a known set of host-state-sensitive tests fails intermittently and passes in isolation.

Observed on `flow:auto #579` (kyle): the first gate run reported **2 failures out of 2859**; an immediate clean re-run of the same suite, with no code change, passed **2859/2859**. The offenders are the documented host-state flakes (registry live filter, unassigned external handle, `group_crud_playwright` launch card).

## Why it matters

A false `fail` stops the run at Step 6 with the branch committed but unpushed, and the recovery is for the model to decide "that looked flaky" and re-run — which is exactly the judgement call a gate exists to remove. It also trains the model to re-run gates on failure generally, which is the wrong instinct: a real failure re-run twice is still a real failure, but a model that has learned "gates are flaky" will narrate past a true red.

The CI side already solved this: ledger entry `0304f02e3c` added `pytest-rerunfailures` with `--reruns 1` on the Woodpecker Playwright steps. The **local** gate runner has no equivalent.

## Proposed fix

Either:

1. **Targeted re-run** — on a non-zero test exit, re-run *only the failed node ids* once (`pytest --last-failed`, or the ids parsed from the report). Pass on the second attempt → report `FLOW_FINISH_GATE: ok` with a `RERUN_PASSED: <ids>` note so the flake is visible and countable, never silent. Fail again → `fail` as today.
2. **Quarantine** — mark the known host-state set with a `@pytest.mark.flaky` / rerun marker, as the CI Playwright steps already do.

(1) is preferable: it needs no per-test annotation, it catches new flakes as well as known ones, and the `RERUN_PASSED` line keeps the flake rate observable instead of hiding it. A re-run that passes should never be reported as a clean run.

## Acceptance

- [ ] A test that fails once and passes on re-run does not stop the flow
- [ ] The gate reports which ids needed a re-run; a reran-and-passed run is never summarized as a plain pass
- [ ] A genuinely failing test still fails the gate after its one re-run
- [ ] Re-run behaviour applies to the local gate only; CI keeps its own policy

Captured by `/self-improvement:retro` from `.claude/friction.jsonl` (kyle), `flow:auto #579`, fingerprint `52e8646b4e`.

[...]

Issue #716, opened 2026-08-13
gh-pr-merge.sh: emit a squash-commit trailer when the PR is up to date with main, to unlock poker-measure's push-pipeline fast path (poker-measure#236)


cooneycw:
## Context

`poker-measure#236` measured that CI throughput on `proxVMpoker04`'s single Woodpecker agent (`WOODPECKER_MAX_WORKFLOWS=2`) is capped hard enough that a merge-heavy wave often has effective PR-verification concurrency of **one** — every squash-merge to `main` spawns a push pipeline that re-runs the full suite and occupies one of the two slots, competing with the next PR's pipeline.

`poker-measure`'s fix (landing in that issue's PR) makes the push-to-`main` test step check the squash commit message for a marker, and if present and VALID, skip the full suite and report success immediately instead of re-running ~10+ minutes of tests — legitimate ONLY when the squashed tree is provably identical to a PR head that already passed `ci/woodpecker/pr/woodpecker`. That provable-identical condition holds if and only if the PR was up to date with `main` at merge time (no intervening `main` commits since the PR's branch point) — checked here because `poker-measure`'s branch-protection ruleset has `strict_required_status_checks_policy: false`, so GitHub does **not** already guarantee this.

## REVISED at the poker-measure#236 gate (2026-08-13): the marker must be self-verifying, not a bare boolean

**The original spec in this issue asked for a boolean trailer (`Woodpecker-Skip-Full-Test: true`) — do NOT implement that. It is a fail-open defect.**

`gh pr merge --squash` composes the squash body from the PR's own commit messages when no explicit `-t`/`-b` is given on that code path. Any commit in any PR whose message happens to contain that literal line propagates the trailer into the squash commit on `main` and skips the full suite — with this script's up-to-date determination never consulted. The safety property lived in a string anyone could type into a commit message, accidentally or otherwise.

### The fix: carry the tested tree hash, not a boolean

1. Before performing the squash merge, determine whether the PR branch is up to date with the target branch (as before — e.g. `git merge-base --is-ancestor <base-tip> <pr-head>`, or the equivalent via `gh pr view --json` fields).
2. If and only if it is, append a git trailer to the squash commit message carrying the **tree hash that was actually tested**:
   `Woodpecker-Tested-Tree: <sha1 of git rev-parse <pr-head-sha>^{tree}>`
3. If the PR is not up to date, do not add the trailer (the push pipeline then runs the full suite as it does today — safe default).

[...]

Issue #621, opened 2026-07-27
lib/cicd: a test step that skips everything exits 0, so the finish gate reports SUCCESS with its acceptance tests never run


cooneycw:
## Friction

Observed on `flow:auto #65` in `agentic-poker` (2026-07-27), a change to a database `UniqueConstraint`.

The `finish` plan reported:

```
Plan 'finish' completed successfully (4/4 steps)
{"success": true, "steps_completed": 4, "steps_total": 4}
```

The test step inside it reported:

```
312 passed, 66 skipped, 1 warning in 55.69s
```

Among those 66 skipped were **the tests that prove the constraint change works**. `agentic-poker`'s ORM suite skips unless `AP_TEST_DATABASE_URL` points at a live Postgres, and the runner sets no such variable. So the gate that exists to answer *"is this safe to merge?"* answered yes without executing the only tests that could have said no.

I caught it only because the repo's `CLAUDE.md` carries a hand-written warning:

> A skipped ORM suite is a passing suite, so check the skip count before trusting a green local run.

That warning is the bug report. A gate whose correctness depends on a human remembering to read the output it just summarised as `SUCCESS` is not a gate.

## Why it happens

`lib/cicd/steps.py:213` decides success on the process exit code alone:

```python
if proc.returncode == 0:
```

`pytest` exits 0 when every test skips. There is no path that inspects the summary line. `mark_step_skipped` / `StepState.SKIPPED` (`state.py:25`, `runner.py:238`) are about *pipeline* steps with `skip_if` — they have nothing to do with a test runner skipping its own tests, and nothing reads that.

## Why now

This is the same class as #617, closed two days ago under exactly the right banner — *"run typecheck in the finish/check gates so a green gate means a green CI."* #617 fixed a step that never **ran**; this is a step that ran and **executed nothing that mattered**. Same promise broken, one layer down.

## Proposed fix

Options, cheapest first:

1. **Surface it.** Parse the summary line and include `skipped` in the step result, so the plan output shows `test: SUCCESS (312 passed, 66 skipped)` instead of a bare `SUCCESS`. Cheap, no policy, and makes the failure legible rather than invisible.
2. **Threshold it.** A configurable `max_skipped` (or `max_skipped_ratio`) per test step in `.claude/cicd.yml`; exceeding it degrades the step to a warning or a failure. Default off so nothing regresses.
[...]

PR #806, opened 2026-09-08
fix(cicd): aggregate every test invocation in a step, and locate security findings

cooneycw:
Two defects in one gate, same shape: **the gate reported things that could not be acted on.** Reported in cooneycw/kyle#838, where the fix turned out not to belong — both halves are here, and kyle needs no change.

> **DO NOT PULL THIS INTO `~/Projects/claude-power-pack` AFTER MERGING.** That checkout is read *live* by every running gate (`flow-finish-gate.sh` resolves `CPP_DIR` to it and runs `uv run --project "$CPP_DIR" python -m lib.cicd`; `lib` is not installed anywhere, and `~/.claude/scripts/flow-finish-gate.sh` is itself a symlink into it). Merging changes nothing for the fleet; the **pull** is the moment the instrument changes under everyone, mid-run. That step is sequenced by the run's orchestrator, when no gate is executing.

## 1. The test count omitted most of the run

`parse_suite_outcome` kept only the **last** summary in a step's output. kyle's `make test` runs pytest twice — non-Playwright, then Playwright — inside one Make target, so only the second invocation's numbers survived.

Measured on kyle, before and after, same tree:

| | reported |
|---|---|
| before | **103** |
| after | **4,218** (`invocations: 2`) |

Under-reported by 97.5%. Exit codes were never affected — `status` is captured from both invocations — so this was a reading defect, not a correctness one. But "103 passed" on a 4,000-test repo either looks like a catastrophic collection failure or gets waved through, and both readings are wrong.

**The previous behaviour was scoped, not an oversight**, and it matters that a reader understands which. Its comment reasoned about *"an early partial"*, and for a partial — or a re-run of a subset — last-wins is correct; summing those would double-count. Two *disjoint* suites is a shape that reasoning did not cover. The #769 failed-id re-run stays correct for a structural reason, checked before proposing the sum: it runs as a **separate `step.execute`** with its own captured output, so it is never in the same text and cannot be absorbed here.

One existing test asserted the old behaviour with the comment *"a `make test` target that runs two suites reports the final one"* — the defect pinned as intent. It is updated deliberately, with the reasoning above recorded in it.

## 2. Summing alone would have left #621's guard blind

**This is why the change is not simply "sum the summaries", and it is what a reviewer would otherwise suggest simplifying it to.**

The #621 warning — the guard against a test step exiting 0 having executed nothing — fires on `nothing_ran`, computed from these same counts. Consider one invocation collecting zero and the other passing 102:

- Old: reported 102, `nothing_ran` False → silent.
- **Summed only: reported 102, `nothing_ran` False → still silent.**
- Now: `invocations: 2`, `empty_invocations: 1` → warned.

An aggregate-only fix would report the honest total, pass a test asserting the honest total, look complete, and leave the guard exactly as blind as it is today — the ticket's own subject reappearing inside its repair. `tests/test_runner.py::test_an_empty_invocation_warns_even_when_the_total_ran` is written to fail against that version, not only against the old one.

## 3. Security findings that could not be triaged

The gate emitted HIGH findings as severity + title only:

```
WARNING: 🟡 HIGH: Hardcoded password in source code   (x5)
```

The location was never missing — `Finding` carries `file_path`/`line_number` and exposes `location`, and the scanners populate both — it was discarded when the message was built. Now:

```
WARNING: 🟡 HIGH: Hardcoded password in source code at core/deploy_checks.py:78 [HARDCODED_PASSWORD]
```

**The located output immediately corrected the ticket's own reading of these findings**, which is the tightest available demonstration of why this matters. The ticket inferred, reasonably, that one finding was `config/settings.py`'s #716 fix being flagged for a literal quoted in its docstring. Running the located gate on the current tree, that finding is **not present at all**, and the one non-test finding is `core/deploy_checks.py:78` — `LEGACY_DB_PASSWORD = "kyle_dev_password"`, the *detector constant* for the `kyle.E003` deploy check that refuses to start on the compose dev password. A known-ignorable, and previously indistinguishable from a real one. The ticket's author could not have known that: they said so, and said the reason was that the findings carry no locations.

`raw_match` is deliberately **excluded**. It may be the secret itself — which is why the model carries `mask_secret` — and these lines land in shared logs, PR bodies and terminal scrollback. Untriageable-but-safe beats triageable-and-leaked; the location is enough to go and look.

## Blast radius, and why it is contained

CPP is the tooling every session's gate runs through, and five workers are gating in parallel on a deadline. The guarantee is the module's own documented contract:

> It is deliberately read-only and advisory: it changes what a step REPORTS, never whether it passed.

`TestVerdictIsIndependentOfCounts` makes that a **demonstrated property rather than an intention**: the same two-invocation output with exit 0 and exit 1 produces SUCCESS and FAILURE respectively, with identical parsed counts in both. If parsing cannot move a verdict, the worst case of a bug here is a wrong number — which is where we already are.

## Verification

- CPP: **2,183 tests pass**, ruff clean, mypy clean on 194 files.
- kyle's **full finish gate** run against this branch via `FLOW_GATE_CPP_DIR`, with the shared checkout untouched: `FLOW_FINISH_GATE: ok`, `"passed": 4218, "invocations": 2, "empty_invocations": 0`, and eight located security findings.
- Every new test observed red first, including the two that must fail against a *correctly-summing* implementation rather than only against today's code.

Reported in cooneycw/kyle#838.

https://claude.ai/code/session_01AWsxq28mDuRZx4QUHqZtrw

### DIFF
--- a/lib/cicd/outcomes.py
+++ b/lib/cicd/outcomes.py
@@ -30,6 +30,15 @@ class SuiteOutcome:
     skipped: int = 0
     errors: int = 0
     framework: str = "unknown"
+    # How many summaries were aggregated into these counts, and how many of
+    # them executed nothing (kyle issue #838). A step whose command runs the
+    # runner more than once - `make test` invoking pytest for two disjoint
+    # suites - produces one summary per invocation, and the TOTAL can look
+    # healthy while one invocation collected zero. Summing without these would
+    # report an honest number and leave #621's guard exactly as blind as the
+    # last-wins behaviour did.
+    invocations: int = 1
+    empty_invocations: int = 0
 
     @property
     def executed(self) -> int:
@@ -45,6 +54,18 @@ class SuiteOutcome:
         nothing at all, so even the skip count is zero).
         """
         return self.executed == 0
+
+    @property
+    def any_invocation_empty(self) -> bool:
+        """True when some invocation executed nothing, even if others did.
+
+        ``nothing_ran`` asks about the total, which is the right question for
+        a single invocation and the wrong one for several: two suites where
+        the first collects zero and the second passes 102 total 102, so the
+        total ran something and the #621 warning stays silent about a half of
+        the gate that proved nothing (kyle issue #838).
+        """
+        return self.empty_invocations > 0
 
     def summary(self) -> str:
         """Human-readable count summary, e.g. ``312 passed, 66 skipped``."""
@@ -64,6 +85,8 @@ class SuiteOutcome:
             "errors": self.errors,
             "executed": self.executed,
             "framework": self.framework,
+            "invocations": self.invocations,
+            "empty_invocations": self.empty_invocations,
         }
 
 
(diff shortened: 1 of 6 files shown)


cooneycw:
## Evidence the fix works, from three independent sources

Landing this standalone ahead of the rest of the gate-semantics batch, because the under-count **actively obstructs the method for detecting stale greens**: the fleet self-audit compares reported test counts to real ones, and a wrong reported count makes a perfectly fresh run look cached.

### The fix working, side by side (projects-ba's audit trail)

Same repo, same shape of run, differing only in which CPP checkout the gate resolved through:

```
kyle-gate-838  [2/4] test: SUCCESS (4218 passed, 1 skipped)   <- PATCHED cpp via FLOW_GATE_CPP_DIR
gate-868       [2/4] test: SUCCESS (103 passed, 0 skipped)    <- shared checkout
gate-868b      [2/4] test: SUCCESS (103 passed, 0 skipped)
gate-868c      [2/4] test: SUCCESS (103 passed, 0 skipped)
```

### Independently reproduced from my own gate logs

Three runs of kyle's suite through the unpatched gate, with the real counts from the same logs:

| run | reported | actually executed |
|---|---|---|
| gate2 | `101 passed` | 3965 + 100 |
| gate4 | `101 passed` | 3965 + 100 |
| gate833b | `103 passed` | 4163 + 102 |

`make test` in kyle runs pytest twice — non-Playwright, then Playwright — and `parse_suite_outcome` kept only the last summary. The reported number is the *second* invocation's, and the first is discarded entirely.

### Why the `empty_invocations` half matters more than the count

Summing alone would have left #621's guard blind: if one invocation collects zero and the other passes 102, the sum is 102, `nothing_ran` is False, and the guard stays silent about a half of the gate that proved nothing. That is the hole #621 exists to close, surviving *inside* the aggregate meant to fix it.

## Gate that produced this green

**`make verify`** — CPP's real gate, not `flow-finish-gate`. That distinction is itself a finding from this batch: `flow-finish-gate`'s Makefile fallback runs **three of these nine targets** and reports `ok`, so every CPP green produced by flow:auto is a true statement about a third of the gate presented as a verdict on the tree. Being filed separately.

All nine confirmed executed, not assumed:

```
ruff check .                    passed
pytest                          2183 passed in 122.93s
mypy .                          no issues in 194 source files
binary-guards                   ok - every shelling-out test is guarded
negative-fixture                ok - every constructed absence asserts its precondition
claude-md-budget                ok - 1187/2000 words
claude-md-links                 ok - every repository-local pointer resolves
claude-md-behavior              ok - obligations remain findable
project-next-vendor             16 files match contract v1.3 at 1724e7d
```

Run from the worktree `claude-power-pack-kyle838-gate-counts`, never the live checkout — that checkout is what every session's `~/.claude/scripts` symlinks resolve into, so editing it changes gates under running gates.

Commits 22480e9 and fc1dfda are unchanged from how they were built and held; I reviewed them and found nothing to correct.

PR #807, opened 2026-09-08
fix(cicd): mark a step whose result was carried from an earlier run

cooneycw:
## The defect

A resumed run reported steps it had **not executed** in exactly the form of steps it had.

Three sessions found this independently, on different days, in different repo states:

| log | line 1 | pytest invocations | marker |
|---|---|---|---|
| `gate846b` (#846, yesterday) | `Resuming ... from step 3` | **0** | `ok` |
| a0's run (today) | `Resuming ... from step 3` | **0** | `ok` |
| `gate4` (today) | `Resuming ... from step 2` | 2 (real) | `ok` |

The first two are the loud version — a green covering a tree that predated the fix, with pytest never invoked at all. **The third is the quiet one, and it is worse for being quiet**: the suite genuinely re-ran and only the *lint* result was stale, for a file the fix had just edited.

All three were caught by knowledge that lives outside the output — grepping line 1 for `Resuming`, counting pytest invocations, remembering a hand-run `make lint`. None of that is a property of the report, and the next person does not have it.

## It is the normal loop, not an edge case

From one session's full audit: **one instance in eleven runs**, appearing exactly when a failed run was retried after a fix.

That is the ordinary workflow. A lint or typecheck failure is almost always repaired by editing a source file — and the tests that would exercise that edit live in the step the resume skips. The more ordinary your workflow, the more likely you hit it.

## The fix

A carried-over step now says so:

```diff
- {"id": "lint", "status": "success", "tests": {...}}
+ {"id": "lint", "status": "success", "executed_in_this_run": false,
+  "carried_from_previous_run": true, "tests": {...}}
```

plus a log line at resume naming the steps that will **not** run, and a second at the end when a *successful* run kept carried results.

### The ordering is the fix, not an implementation detail

`step_details` was previously emitted **only on failure**, and the obvious repair — read the state back in `run_plan` and annotate it — is a **silent no-op on the green path**, because a successful run deletes its state file. I wrote that version first; the end-to-end test is what caught it.

So the detail is captured *before* cleanup and carried on the result. The old guard withheld it precisely where it was needed: **a red run is already read carefully, a green one is not.** A test pins the constraint (`test_the_state_file_is_gone_on_success_so_detail_must_be_captured`) so the capture is not later reverted as redundant — without it, the next person removes it and the feature quietly stops working on the success path while every unit test still passes.

## This is half of a pair — please don't merge one and consider it closed

Neither half alone makes the summary trustworthy:

- **cpp#806** fixed the **count**. `parse_suite_outcome` kept only the last summary, so `make test`'s two pytest invocations reported the second as the whole run: `103 passed` against a real 4,282 + 102. That was wrong on *every* run, cached or fresh.
- **This PR** fixes whether the step **ran at all**.

A reader needs both to tell a full run from a partial one, and either from a remembered one.

## Tests

Nine, split across the two halves they pin — the summary rendering (`test_runner_state.py`) and the runner that feeds it (`test_runner.py`). A summary that can mark correctly, given a number nobody supplies, marks nothing.

Confirmed to **fail without the fix** (4 of 5 in the state suite; the fifth is the backward-compatibility guard rail, correctly passing either way — omitting `executed_from` must keep the old shape exactly, so the marking is additive and never a silent rewrite).

Both directions are pinned: a carried step is marked, a step executed *this* invocation is **not**, and a pending/skipped record before the resume point is not called "carried" — it has no result, so claiming one would be a false statement in the direction of more confidence.

## Gate

**`make verify`** — CPP's real nine-target gate, not `flow-finish-gate`, which runs three of the nine. All nine confirmed *executed* rather than inferred from exit 0:

```
ruff check .          passed          binary-guards         ok
pytest                2192 passed     negative-fixture      ok
mypy .                194 files       claude-md-budget      ok - 1187/2000 words
                                      claude-md-links       ok
                                      claude-md-behavior    ok
                                      project-next-vendor   16 files, contract v1.3
```

Run from a worktree, never the live checkout — that checkout is what every session's `~/.claude/scripts` resolves into, so editing it changes gates under running gates.

https://claude.ai/code/session_01S7zji5qgFeVaffLwDKRMXp

### DIFF
--- a/lib/cicd/runner.py
+++ b/lib/cicd/runner.py
@@ -26,7 +26,7 @@ from pathlib import Path
 from typing import Any, Optional, TextIO
 
 from .outcomes import parse_failed_node_ids
-from .state import RunState
+from .state import RunState, StepStatus
 from .steps import GATE_STEP_IDS, ShellStep, StepDef, get_plan_steps
 
 # Variables the runner launcher injects (or a parent venv leaks) that must NOT
@@ -133,6 +133,17 @@ class RunResult:
     # re-run that passed is a DIFFERENT qualification and must not borrow #621's
     # sentence.
     reruns: list[dict[str, Any]] = field(default_factory=list)
+    # Step ids whose result was earned in an EARLIER invocation and carried
+    # into this one by a resume (issue #838 follow-up). Its own channel and not
+    # a `warning`: nothing is wrong, but the reader must be able to tell a
+    # result just earned from one merely remembered. Captured on the RESULT
+    # rather than read back from state, because the state file is deleted on
+    # success - which is exactly why the old `step_details` block could never
+    # have shown this on the green path it matters most on.
+    carried_from_previous_run: list[str] = field(default_factory=list)
+    # Full per-step detail, captured BEFORE the success cleanup removes the
+    # state file so a green run can report it at all.
+    step_details: list[dict[str, Any]] = field(default_factory=list)
     # Step ids that skip_if-skipped this run (issue #628). A skipped GATE step
     # (lint/test/typecheck) means the gate verified nothing about the change, so
     # flow-finish-gate.sh reads this to report `warn` and NAME the skipped gates
@@ -192,6 +203,7 @@ class DeterministicRunner:
         existing = RunState.find_latest(plan_name, self.project_root)
         if existing:
             self._log(f"Resuming failed run {existing.run_id} from step {existing.current_index + 1}")
+            self._warn_carried_over(existing)
             return self._execute(existing, step_defs)
 
         # Load step definitions
@@ -225,6 +237,7 @@ class DeterministicRunner:
             )
 
         self._log(f"Resuming run {run_id} from step {state.current_index + 1}")
+        self._warn_carried_over(state)
         # Reset state to running for resume
         state.status = "running"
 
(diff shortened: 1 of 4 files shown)


cooneycw:
## The defect underneath this one: the run state has no expiry and no scope

projects-95 reproduced the resume in kyle's main checkout and found the run state is not stale by minutes:

```
Resuming failed run finish-2c7ab2e9 from step 2
.claude/runs/finish-2c7ab2e9.json   last written  2026-08-31 14:30
```

**Eight days old.** Every gate invocation in that checkout since August 31 has been resuming the *same run id*, caching whatever that run had already completed — across unrelated branches, unrelated issues, unrelated people.

### What that means for this PR

It reframes the fix rather than invalidating it, and the distinction is worth being precise about:

**This change makes the problem visible, not absent.** A perfectly honest  would have been telling people — accurately — that their gate was reusing results earned in *August*. That is a far better failure than the silent one it replaces, and it is still not a working gate.

So there is a lifecycle defect beneath the display defect: a run id that survives eight days and every branch switch. **I am not fixing that here**, deliberately — it is a scope-and-expiry question (should state be keyed to a branch? a tree hash? expire after N hours? refuse to resume across a HEAD change?) that wants its own decision, and bundling it would make this change unreviewable. Filing it separately.

Worth noting the two interact in the right direction: once carried steps are marked, the eight-day case becomes *diagnosable* by reading the output, which is how anyone would notice the lifecycle bug in the first place.

## Two independent defects, confirmed independent

projects-95 separated what had been treated as one thing, and their evidence settles it:

| | defect | fix |
|---|---|---|
| 1 | resume from stale run state | **this PR** (display) + the lifecycle question above |
| 2 | under-count on **fresh** runs | **cpp#806**, merged and now pulled live |

Their #859 gates were unambiguously fresh — , no  line, and  statuses on the failing one, which only a from-scratch run produces — and *still* reported  against a real 4,284.

So the under-count was never a caching artefact, and this PR does not close it. Neither closes the other.

cooneycw:
**Correction to the comment above** — three inline code spans were lost to shell quoting when I posted it. Restoring them, since two carry the actual content:

1. "A perfectly honest ___ would have been telling people" should read: a perfectly honest **`carried_from_previous_run: true`** would have been telling people, accurately, that their gate was reusing results earned in August.

2. The evidence that the two defects are independent: projects-95's #859 gates were unambiguously **fresh** — `Plan 'finish' completed successfully (4/4 steps)`, no `Resuming` line, and `pending` statuses on the failing one, which only a from-scratch run produces — and still reported `"executed": 103` against a real **4,284**.

So the under-count was never a caching artefact. This PR does not close it, and cpp#806 does not close this one.

Apologies for the noise; the substance above is otherwise as intended.

cooneycw:
Closing and reopening to re-trigger CI. The PR pipeline (1327) has been stuck in `created` since the push, was passed over by newer pipelines, and an explicit restart produced an errored 1330. The original push pipeline (1326) failed because its **clone step failed after 60s**, which killed the validate step mid-run at 55s — not a test failure. Infrastructure has since recovered (1328 SUCCESS on a sibling branch), so a fresh PR event should get a clean pipeline. No code change.

cooneycw:
## Post-merge: the defect is worse than this PR described — resumes COMPOUND

Measured by projects-a0 on a different repo, and it belongs in this PR's record as the headline rather than as a detail. Two resumes fired **back to back**:

```
10:41:34   resumed from step 2   tests DID run (4181+102). Lint did not.   looked normal
10:42:29   resumed from step 3   neither lint NOR tests ran.               FLOW_FINISH_GATE: ok
```

**Fifty-five seconds apart, against a ~9-minute full gate.**

So the failure is not "a resume can skip the tests". It is that **resumes compound**, and the *first* skip is camouflaged precisely because a partial resume still prints plausible test counts — `4181+102` is a real, complete-looking number. Nothing about that line invites suspicion. Only the *second* resume, standing on the first, skipped everything.

### Why this is the argument for the change rather than a footnote

The only reliable tell was the **clock** — a green in under a minute is prima facie evidence that nothing ran. And even then it took grepping line 1 for `Resuming` to confirm, because `[3/4]` and `[4/4]` read perfectly normally.

**A human noticing "that was too fast" is not a detection mechanism.** It depends on knowing the expected duration, watching the timestamps, and being suspicious at the right moment. Three sessions found this defect independently today and all three needed knowledge that lives outside the output: grepping line 1, counting pytest invocations, remembering a hand-run `make lint`, reading a clock.

`carried_from_previous_run: true` is the same fact, present in the artefact, requiring none of that.

### What this does not fix

Still true and still separate: the run state has **no expiry and no scope** — an eight-day-old run id was being resumed across unrelated branches. This marking makes that *visible*, not absent. A perfectly honest `carried_from_previous_run: true` would have accurately reported results earned in August. The lifecycle question is unfixed and unfiled as of this comment.

Issue #821, opened 2026-09-11
flow-wave-mailbox: count_watchers cannot distinguish a live watcher from an orphaned subshell of a dead one (identity defect, measured 12/12)


cooneycw:
`count_watchers()` cannot distinguish a live watcher from an **orphaned command-substitution subshell of a dead one**. This is an identity defect, not a race: no retry, timeout increase, or hardening of the ancestor walk fixes it.

Surfaces today as `tests/test_flow_wave_mailbox.py::TestWatchStatus` failing intermittently - including in a bare `-k TestWatchStatus` run - most often as `test_one_watcher_counts_once_despite_its_own_subshells` asserting `watch never armed`.

## Measured, not inferred

Scan `/proc` for the watcher argv **before** each run, then run the failing test alone:

| run | orphans before run | result |
|-----|--------------------|--------|
| 1-4 | 0 | PASS |
| 5 | 4 | **FAIL** |
| 6 | 5 | **FAIL** |
| 7-12 | 0 | PASS |

**Ten passes at zero orphans, two failures at four and five. No exceptions in twelve runs.**

## The mechanism

`watcher.kill()` followed by `communicate()` **does** reap the direct child - that process is genuinely gone. What survives is its **command-substitution subshells**. They are forked, not exec'd, so they carry byte-identical argv; SIGKILL on the parent never reaches them; they reparent to PID 1 and live until their own `--timeout 30` expires. One killed watcher leaves **four to five** survivors.

To `count_watchers`, that is a crowd of live watchers on the right role and the right wave.

The consequence for the test: every `TestWatchStatus` case shares `WAVE = "testwave"` and role `"1"` (`tests/test_flow_wave_mailbox.py:60`). The duplicate-arm guard is a **one-shot** read at `scripts/flow-wave-mailbox.sh:1024`, before the poll loop. So the next watcher's single guard read sees the orphans, refuses with `exit 4`, and **never stamps a heartbeat** - and `_live_watcher` waits its full 15-second deadline for a heartbeat that will never arrive.

### Why Phase 2 cannot catch it

`watcher_roles_proc()` Phase 2 drops any match whose **parent also matched**, and its comment states the invariant it relies on:

> *"A real watcher's parent is its launcher (a `-c` wrapper or a shell), which never matches; a subshell's parent is always the watcher."*

[...]

PR #825, opened 2026-09-11
fix(flow-wave-mailbox): acknowledge receipt explicitly (Closes #815)

### LINKED ISSUE #815: flow-wave-mailbox: acknowledge receipt explicitly; cursor-on-output hides unseen mail and failed output copies
`watch --consume` and ordinary mailbox reads advance the read cursor when output is emitted, before the conversational recipient confirms receipt. If a tool response is dropped or only part of a batch reaches the agent, the next normal read reports empty. The append-only message history still exists, but normal delivery no longer offers those messages.

The `--out` safeguard also advances cursors before the destination write succeeds, so its claimed durable-copy-before-consumption ordering is not enforced.

## Verified evidence and reproduction

Source: CPP `origin/main` at `fbeb0245e85a4e41b51072986ab53cae59ae29b1`, verified against remote main on 2026-09-10.

- [drain_role](https://github.com/cooneycw/claude-power-pack/blob/fbeb0245e85a4e41b51072986ab53cae59ae29b1/scripts/flow-wave-mailbox.sh#L468) prints each box and sets its cursor to the highest revision.
- [read --out](https://github.com/cooneycw/claude-power-pack/blob/fbeb0245e85a4e41b51072986ab53cae59ae29b1/scripts/flow-wave-mailbox.sh#L943) captures `drain_role` output, which has already advanced cursors, then attempts the file write. The `--from` path has the same ordering at lines 926-930.

Isolated checks used the exact script with a fresh temporary `FLOW_WAVE_MAILBOX_DIR`, synthetic messages and no live mailbox access:

1. Enqueue two messages and assert unread=2. Run `watch --role worker --wave fixture --timeout 0 --consume`, capturing but withholding its output from a simulated recipient. Exit=0, unread=0, and the next ordinary read reports empty. Both bodies were in the captured response: the failure boundary is after helper output, not a claim that the helper itself printed only one message.
2. Positive control: `watch --peek` returns the body and leaves unread=1.
3. Enqueue a fresh message and assert unread=1. Run `read --role worker --wave fixture --out <existing-directory>`. The destination write fails with exit 2 and `cannot write --out`, but unread becomes 0 and no durable output copy was created.

## Requested change

Introduce explicit receipt acknowledgement tied to stable message identities (including wave, box and revision). Receiving or surfacing output must leave an unacknowledged message recoverable through the normal delivery path. Separate stored, output/surface attempted, recipient acknowledged, and any later task answer; acknowledgement is neither task completion nor permission to proceed.

Define compatible migration from existing cursors without silently declaring historical mail acknowledged. An explicit legacy consuming mode can remain if clearly identified; the normal agent workflow must use the new receipt contract. Do not promise exactly-once task execution.

## Acceptance and verification

- [ ] A dropped tool response, interrupted read, truncated batch or listener death before acknowledgement leaves every unacknowledged message available for replay with its original identity.
- [ ] Acknowledging a later message cannot acknowledge an unseen earlier message. Bind acknowledgement to exact received identities; out-of-order and partial batch receipt must be safe.
- [ ] Repeated acknowledgements and replay are idempotent. Concurrent send/receive/ack does not skip revisions, overwrite mail, or create ambiguous duplicate deliveries.
- [ ] If `--out` is retained as a durable-copy operation, a destination open/write failure must not consume or acknowledge messages. Exercise worker, orchestrator-wide, and `--from` paths with asserted failing destinations and a successful writable control.
- [ ] Sender status distinguishes storage, attempted surfacing and explicit receipt, with age for unacknowledged messages. Offline queueing remains successful; acknowledgement is not shown as task completion. Response correlation must not be inferred merely from acknowledgement.
- [ ] Update worker/orchestrator command sources and generated skill mirrors together. Integration tests cover actual command invocation and multi-box/batched output, including partial acknowledgement and restart.

## Existing work and boundaries

[CPP issue 801](https://github.com/cooneycw/claude-power-pack/issues/801) already identified consume-before-receipt, but merged [PR 803](https://github.com/cooneycw/claude-power-pack/pull/803) changed watcher status. [PR 796](https://github.com/cooneycw/claude-power-pack/pull/796), associated with [issue 792](https://github.com/cooneycw/claude-power-pack/issues/792), introduced explicit consume/peek choice and `--out`; neither is a recipient acknowledgement protocol.

Coordinate with [CPP issue 814](https://github.com/cooneycw/claude-power-pack/issues/814) for listener recovery; storage/receipt semantics can be implemented independently of its runtime adapter. The existing history can be inspected for manual recovery, so this is loss of normal delivery visibility, not proven deletion of stored bodies. Kyle already has explicit durable acknowledgement and is not the owner of this defect.

cooneycw:
## Summary

`read`/`watch` advanced a read cursor as a side effect of PRINTING output,
conflating "surfaced" with "received". A dropped tool response, a truncated
batch, or a `--out` destination that failed to write all silently and
permanently lost the message - the cursor had already moved past it, so the
next ordinary read reported empty while the caller had never received
anything. `read --out <existing-directory>` reproduced this directly: the
write fails (exit 2) but the cursor still advanced.

- `.cursor-<box>` is retired for `.ack-<box>`, a SET of explicitly
  acknowledged revs (not a watermark) - acking rev 7 must not silently imply
  5 and 6 were also received.
- New `ack --role <role> [--from <role> | --box NAME] (--revs R[,R...] |
  --all-unacked)` verb - the only thing that ever records a receipt. Refuses
  (exit 2, all-or-nothing on a batch) a rev that does not exist yet in the
  box.
- `read --peek` / `watch --peek` print WITHOUT acknowledging anything - a
  dropped response after a peek leaves the message recoverable through the
  very next ordinary read or watch.
- `read` (bare) / `watch --consume` keep the pre-#815 print-and-acknowledge
  behavior as a named legacy convenience (the issue's explicit allowance),
  unchanged in timing - prefer `--peek` + a later `ack` for anything a
  dropped response must not lose.
- `read --out` now writes its destination BEFORE any acknowledgement - the
  exact ordering the issue's evidence reproduced.
- `list`'s `CURSOR` column/JSON key becomes `ACKED`/`acked`. Required a
  companion rename in `flow-wave-registry.sh`, which joins the mailbox's own
  `list --json` rather than reimplementing the parse.
- No migration tooling needed: wave dirs are ephemeral (`$XDG_RUNTIME_DIR`,
  wiped at reboot), so a box predating this fix simply has no `.ack-<box>`
  file - absence means "nothing acknowledged", never an implicit claim.
  `ack --all-unacked` is the explicit way to close that gap on purpose.
- `make codex-skills` regenerated the affected bundled copies.

## Test plan

- 22 new tests in `tests/test_flow_wave_mailbox.py`: peek-never-acks is the
  core regression pin (a dropped response still replays through normal
  delivery); explicit ack after a peek stops the replay; legacy auto-ack
  preserved; exact-identity binding (partial/out-of-order/idempotent/refuses
  an unseen future rev/all-or-nothing batch/`--all-unacked`); orchestrator
  multi-box disambiguation; `--out` write-before-ack (including the exact
  failure-mode reproduction from the issue); `list` reports `acked`
  distinct from `unread`, durable across processes; migration is
  conservative (no `.ack-<box>` file = everything unacknowledged).
- One assertion updated in `tests/test_flow_wave_registry.py` for the
  `cursor` -> `acked` JSON rename.
- All 73 pre-existing `test_flow_wave_mailbox.py` tests pass UNMODIFIED -
  the CLI contract (`--peek`/`--consume` timing, `--from`, `--out` refusal)
  is preserved exactly.
- `make verify` green: lint, 2481 tests, mypy, binary-guards,
  negative-fixture, claude-md-*, project-next-vendor.
- The deployed `flow-finish-gate.sh` itself, run against this worktree,
  green - first line `Starting plan`, not `Resuming`.

Two full-suite runs of the wave test files together surfaced intermittent
failures in `TestWatchStatus`/`TestDuplicateWatchers`/`TestWatchIsTheWake` -
a different test each time, never reproducing alone. This is issue #821
(the orphaned-subshell watcher-count flake, already filed and independently
confirmed pre-existing): this diff never touches `watcher_roles_proc`,
`self_chain_proc`, `count_watchers`, or the duplicate-arm guard, confirmed
hunk-by-hunk, so there is no causal path from this change to that
mechanism. A clean 2481-test `make test` run is included as corroboration,
not as the sole proof.

Closes #815


cooneycw:
## The contract validated itself against this very PR's own review, live

While reviewing this PR, three delivery-loss incidents happened on the fleet's own substrate mailbox (a separate, independently-built client - not this script - but the same non-consuming-read contract this PR argues for), and all three recovered for the same reason:

1. A watch backgrounded with `&` inside a tool call had its output redirected to `/dev/null` - a delivery was lost. `read` returned it again, same id, as a fresh delivery, because reading never consumes.
2. A watch process died with an unexplained exit 144 and produced no output. Two messages came back unacknowledged and were recovered by the next `read`.
3. A re-arm was simply forgotten for ~25 minutes while seven messages queued, three from `master:kyle`. All seven were recovered by `read` once the mistake was noticed.

None of these are hypothetical failure modes this PR guards against speculatively - they are exactly the "dropped tool response" / "forgot to act on surfaced output" class this PR's core regression pin (`test_peek_never_acknowledges_so_a_dropped_response_still_replays`) exists for, reproduced independently by the two sessions who designed this fix, on a sibling system, within the same working session.

## A limitation worth recording, surfaced by incident 3's aftermath

The substrate client hit has no `--all` equivalent: `read` there returns *unacknowledged* mail only, so acking a batch before reading every body in it makes those bodies unretrievable afterward - ack is effectively terminal for retrieval on that client.

`flow-wave-mailbox.sh`'s `read --all` does not have this limitation: it re-shows a box's FULL history, already-acked messages included, so acking here never forecloses later re-reading a message's content for audit or reference - only `unacked_revs_in`/`extract_unacked` (the default, unacked-only view) change based on ack state. Noting this explicitly since it was a live design question raised during review, and the answer for this script is: yes, `--all` already preserves that replay, by design (see the header's ACKNOWLEDGEMENT section - `--all` is documented as "a display request, not a receipt").

Issue #801, opened 2026-09-07
flow-wave-mailbox: watch --status reports 'armed' for a dead watch; three sessions went silently deaf in one wave


cooneycw:
Three sessions in one wave went silently deaf today, each for tens of minutes,
each while the roster read healthy. The wave's own docs make the mailbox watch
load-bearing, and the watch has three independent failure modes that all present
as "fine".

Evidence from a single live wave (`docker-list`, kyle, 2026-09-07). Every item
below was reproduced, not inferred.

## 1. `--status` reports `armed` for a watch that has already exited

```
role 'orchestrator' wave 'docker-list': last wake handled 2s ago (state: armed);
0 live watcher process(es); re-armed: no
FLOW_MAILBOX_WATCH_STATE=armed
FLOW_MAILBOX_WATCHER_COUNT=0
```

`WATCH_STATE` is a heartbeat STAMP — only as fresh as the last wake — so it reads
`armed` for a watch that is not running. `WATCHER_COUNT` is the field that
answers "is anything watching right now", and the two disagree precisely when it
matters.

This is not theoretical: the wave orchestrator broadcast "check with `--status`;
if it is still holding, do nothing" to three workers, based on the state word.
A worker caught the error and reported it. Following that instruction literally
would have left all three deaf with a green-looking status.

## 2. The watch is one-shot, so a missed re-arm is permanent and silent

It fires on mail and exits. A role that does not re-arm never hears anything
again, and nothing surfaces that. In this wave:

- worker-C went deaf for ~50 minutes holding a Step-3 gate request; the
  orchestrator's ruling sat unread the whole time. Found only because a human
  asked "what about worker C".
- worker-C went deaf a **second** time immediately after reporting a PR done —
  its next assignment landed in the gap. It self-reported "standing by idle",
  which was wrong when sent, because it could not see the assignment it had.
- worker-A sat BLOCKED on a lane ruling that had already been written.
- The orchestrator's own watch died repeatedly, twice leaving a worker blocked
  on a ruling it had already made.

## 3. `--consume` marks mail read on a wake nobody reads

Arming with `--consume` advances the cursor when the watch fires. If the process
that armed it is mid-turn and never reads the box, the mail is marked read and
never seen. The orchestrator lost a worker's message this way and recovered it
only because the boxes are append-only (`awk` over `inbox-<role>.md`).

`--peek` is the documented alternative and has the opposite failure: it does not
[...]

Issue #828, opened 2026-09-11
install-drift: the helper comparison cannot see a MISSING helper - it iterates the installed side only, and reported ok while one was absent


cooneycw:
`install-drift.sh`'s helper comparison iterates the **installed** directory, so a helper that exists in the checkout but is **not installed** is structurally invisible to it. The script reports `ok` while the drift is live.

This is the mirror image of the membership floor #823/#826 added for `~/.codex/skills/`: that floor makes an *absent tree* report `unknown` rather than `clean`. The helper loop needs the same property in the other direction - an *absent install* of a helper that the repo ships is drift, not silence.

## The code

`scripts/install-drift.sh:125-133`:

```bash
for installed in "$SCRIPTS_DIR"/*.sh; do
    ...
    [ -e "$installed" ] || continue
    base="$(basename "$installed")"
    source_helper="$CHECKOUT/scripts/$base"
    [ -f "$source_helper" ] || continue     # installed-but-not-in-repo: correctly skipped
    if cmp -s "$source_helper" "$installed"; then
```

Every path starts from something already present in `$SCRIPTS_DIR`. There is no iteration over `$CHECKOUT/scripts/*.sh`, so "the repo ships a helper nobody installed" is not a state the loop can reach.

## It happened today, and the detector said `ok` throughout

`~/.claude/scripts/delegated-run-check.sh` was missing on this host. It was found when a worker invoked it and got **exit 127**, then fell back to the in-checkout copy - not by the instrument built for this class.

`install-drift.sh` runs at every session start (`hook-pending-retro.sh` invokes it; see its lines 21-23). Its output while the helper was missing would have been what it prints now that the gap is closed:

```
install-drift: checkout /home/cooneycw/Projects/claude-power-pack
  host helpers       /home/cooneycw/.claude/scripts
    36 current, 0 stale

install-drift: installed helpers match the checkout.
INSTALL_DRIFT: ok
```

With 35 of 36 installed it would have read `35 current, 0 stale` and still `ok` - a confident all-clear, differing only in a count nobody has a baseline for.

**This is #622's own failure mode surviving inside #622's fix.** The issue was *"sessions silently execute weeks-old command text"*; this is *"sessions cannot execute helper code at all, and the check says everything matches."*

## Requested change

Iterate the **checkout** side as well, and report a repo helper with no installed counterpart as its own state.

Suggested vocabulary, matching what #826 established next door:

- **stale** - installed, content differs (existing behaviour, unchanged)
[...]

Issue #816, opened 2026-09-11
commands: PYTHONPATH points one level too deep, so every documented `python -m lib.*` invocation fails outside the CPP root


cooneycw:
The command documents tell the reader to run CPP's Python entry points with `PYTHONPATH` pointed one level too deep, at `$CPP_DIR/lib` instead of `$CPP_DIR`. `lib` is the package, so its **parent** belongs on the path. The documented form only works in the one directory where an unrelated accident rescues it, and that directory is the one nobody is supposed to be in.

This is not a new defect. Issue #430 established the correct contract and it is stated plainly in the tree today - `.claude/commands/cicd/verify.md:42` ("`PYTHONPATH` pointed at `$CPP_DIR` - the **parent** of `lib/`"), `.claude/commands/flow/deploy.md:126`, `.claude/commands/flow/auto.md:1085`. **The migration was applied to 13 sites and missed 57.** The partial migration is the bug.

## Reproduction

Verified on `origin/main` at `fbeb024`, 2026-09-11.

```
$ cd ~/Projects/claude-power-pack
$ PYTHONPATH="$CPP_DIR/lib" python3 -m lib.security gate flow_finish
Security gate passed.                                   # exit 0

$ cd /tmp
$ PYTHONPATH="$CPP_DIR/lib" python3 -m lib.security gate flow_finish
Error while finding module specification for 'lib.security'
(ModuleNotFoundError: No module named 'lib')            # exit 1

$ cd /tmp
$ PYTHONPATH="$CPP_DIR" python3 -m lib.security gate flow_finish
Security gate 'flow_finish' FAILED.                     # exit 1, but it RAN
```

`-m` puts the current working directory on `sys.path`. From the CPP root, `lib` resolves out of the cwd and `PYTHONPATH` contributes nothing at all. From anywhere else there is no `lib` on the path and the import fails.

## Naive verification misleads in BOTH directions

**The broken form looks fixed.** The text says `$CPP_DIR` precisely because the reader is expected to be in another project, which is the case that fails - but anyone checking from inside the CPP checkout gets a clean green from the broken form.

**The fixed form looks broken.** In the third case above the gate ran correctly and then reported a HIGH finding, because `lib.security gate` scans the working directory and `/tmp` held unrelated scratch files. That exit 1 is the tool working.

[...]

PR #830, opened 2026-09-11
fix(flow-wave): restore mailbox notification without conversational re-arming (Closes #814)

### LINKED ISSUE #814: flow-wave: restore mailbox notification after watcher exit without conversational re-arming
When a wave mailbox watch delivers a message or times out, it exits. The agent must launch another background call before it can hear the next message. A missed completion notification or forgotten re-arm leaves persisted assignments waiting indefinitely. Current status reporting detects a dead watch, but does not restore delivery to the agent.

## Verified evidence

Source: CPP `origin/main` at `fbeb0245e85a4e41b51072986ab53cae59ae29b1`, checked against the remote main ref on 2026-09-10.

- [Watch loop](https://github.com/cooneycw/claude-power-pack/blob/fbeb0245e85a4e41b51072986ab53cae59ae29b1/scripts/flow-wave-mailbox.sh#L1031): delivery drains mail and exits 0; empty expiry exits 5. Neither starts a successor listener.
- [Worker protocol](https://github.com/cooneycw/claude-power-pack/blob/fbeb0245e85a4e41b51072986ab53cae59ae29b1/.claude/commands/flow/register.md#L302) assigns re-arming after every wake to the conversational agent. The orchestrator has the same responsibility.
- [Send result](https://github.com/cooneycw/claude-power-pack/blob/fbeb0245e85a4e41b51072986ab53cae59ae29b1/scripts/flow-wave-mailbox.sh#L884) reports storage success without the recipient's current notification-route health. Separate `watch --status` already distinguishes armed, stale, dead, absent and unknown.

Isolated reproduction using this exact script and a fresh `FLOW_WAVE_MAILBOX_DIR`: send two messages; `watch --role worker --timeout 0 --consume` returns 0; send a third message; inventory reports one unread message and `watch --status` reports dead / zero watchers. An empty watch with timeout 0 returns 5 and likewise leaves no listener. These prove the lifecycle gap, not a particular external harness failure.

## Requested change

Make continued listening and notification recovery an explicit runtime responsibility, with one owner per wave/role and shutdown on role release or wave completion. Evaluate the existing preference for verified direct session messaging before adding another daemon. A fallback must have a demonstrable path from stored message to agent notification and receipt.

Keeping a shell alive with `--follow` is insufficient: the current protocol depends on command completion to resume the agent. Likewise a heartbeat proves polling, not that the harness will deliver its output. Expose route readiness, last observed exit/failure and unconfirmed notifications separately from process liveness. Surface this through sender-visible status/send results while preserving valid offline enqueue.

## Acceptance and verification

- [ ] An independently running idle agent receives and acknowledges two sequential messages without a human nudge or conversational re-arm after each one. Repeat while it is busy with a bounded repository task.
- [ ] Exercise normal delivery, empty timeout, invalid startup arguments, a removed launch cwd, externally terminated listener, and a dropped harness completion notification. Recover within a documented bound or report the route as unavailable; never report confirmed receipt from shell output alone.
- [ ] Use a stable launch directory and retain sanitized failure evidence. Invalid invocations must not restart in a tight loop; duplicate owners and released roles must not accumulate listeners.
- [ ] A positive control with a live polling process but disconnected notification path fails the readiness/receipt check. Offline enqueue succeeds and eventual reconnection delivers pending messages.
- [ ] A busy recipient or unanswered task is not classified as a dead agent and does not trigger agent restart. Unknown route state remains unknown.
- [ ] Tests drive registration/notification integration, not only the polling helper. Record supported harness/backend and actual agent receipt evidence separately from mocked tests.

## Existing work and boundaries

[CPP issue 801](https://github.com/cooneycw/claude-power-pack/issues/801) described this residual, but [PR 803](https://github.com/cooneycw/claude-power-pack/pull/803) fixed status fusion, not automatic recovery. [Issue 792](https://github.com/cooneycw/claude-power-pack/issues/792) and [PR 796](https://github.com/cooneycw/claude-power-pack/pull/796) already cover explicit consume/peek selection, watcher counting, duplicate checks and read safeguards. Preserve those fixes.

Explicit message acknowledgement/replay is owned by [CPP issue 815](https://github.com/cooneycw/claude-power-pack/issues/815); coordinate the notification handoff with it. No dependency on Kyle implementation or new cross-host file sharing. Host-local CPP scope remains host-local. Historical claims about memory-driven harness kills are unverified here; recovery must work regardless of termination cause.

cooneycw:
## Summary

`watch` delivers one message (or a timeout) and exits by design, so every wake needs a NEW `watch`, armed by whoever is listening - and "whoever" was always the conversational agent. Three sessions hit this building #814 and #815 themselves in ~90 minutes (see the incident comment on issue #815): a watch backgrounded with `&` into `/dev/null`, a watch that died with no output, and a 25-minute forgotten re-arm that left seven messages waiting, caught only because a different session noticed the silence from outside.

**No code in this repo can make the harness re-invoke a specific agent's conversation on a background process's completion** - that wiring belongs to the harness. This PR does not claim to close that gap. It makes a listener's existence independent of anyone remembering to arm one, and makes a silent failure of the remaining gap observable with a bound instead of invisible forever.

Three deliverables:

1. **`supervise --role <role> [--wave W] [--timeout SEC] [--interval SEC]`** - the invocation returns promptly with its own verdict (`supervising` or `duplicate`, exit 4); a DETACHED daemon it forks repeatedly re-arms `watch --consume`, so each inner watch still triggers the harness's notification for whoever is listening. Single ownership is a **lifetime `flock`**, not a PID file - a PID-file guard was the first draft and was corrected in review: PIDs are reused by the kernel, so a dead daemon's recycled PID would make `kill -0` report "alive" and reproduce #821's argv-collision defect one layer up. Shutdown polls the registry and exits on `FLOW_WAVE_LIVENESS=released` (not on `free` alone, which also covers "never registered" - a legitimate standalone use that must not self-terminate). Argument validation once at start, exponential capped backoff on a crashed inner watch, sanitized evidence log (timestamps and exit codes only, never message bodies).
2. **`route_state()`** - answers "has anything been acknowledged" (#821-immune, built on #815's ack evidence) rather than "is a process polling" (#821-tainted). Four states, not three: an empty box must never read as `confirmed` - `confirmed` / `pending` / `unconfirmed` (older than `FLOW_WAVE_ROUTE_UNCONFIRMED_SECS`, default 900s) / `unknown`.
3. **Sender-visible reporting** - `list`'s `routes` array is a sibling of `watches`, never merged into it (the #801 lesson enforced in the schema); `flow-wave-registry.sh` gets a matching join, visible from a roster sweep.

`register.md`/`wave.md` Step 4 now launch `supervise` by default; a bare, manually re-armed `watch` remains documented as the fallback. Opt-in was considered and rejected: the sessions who most need a supervisor are precisely the ones who will forget to reach for it.

## Test plan

17 new tests in `tests/test_flow_wave_mailbox.py`:
- Route readiness: all four states, the bound is env-overridable without sleeping, orchestrator multi-box takes the worst state, JSON schema keeps `routes` a sibling of `watches`.
- Supervise: returns promptly with its own verdict; a message sent after arming is delivered and acknowledged with **no further command from the test**; a second supervise is refused while the first lives; ownership is the **lock**, not the pidfile (pinned by corrupting the pidfile and confirming nothing changes); the lock releases and a new supervise succeeds once the daemon dies; the evidence log never carries a message body; a killed inner watch gets a replacement after backoff; an invalid role exits immediately with no restart; shutdown on role release via the real registry script; an unreadable registry sibling fails open.

`make verify` green: lint, 2737 tests, mypy, binary-guards, negative-fixture, claude-md-*, project-next-vendor. The deployed `flow-finish-gate.sh` itself, run against this worktree, green - first line `Starting plan`, not `Resuming`.

Not automated, and said so rather than claimed: the issue's own acceptance criterion of "an independently running idle agent receives two sequential messages without a human nudge" is a live-harness property no pytest suite can prove - a documented manual/fleet-level verification step, the way this run's own incidents (recorded on #815) proved and disproved delivery properties for a sibling system in practice.

Closes #814


cooneycw:
## Bug 2 is the day's pattern, inside the fix for the day's pattern

Worth naming explicitly, because it wasn't caught by review - it was caught by a failing test, and it is not a coincidence which one failed.

My first shutdown-check treated `flow-wave-registry.sh get`'s `free` verdict as "this role has been released, shut down." But `free` means two distinct things collapsed into one word: **"released"** and **"never registered at all."** Treating absence of a registration as evidence of a release killed the daemon on its very first loop check, whenever `supervise` was used standalone (a legitimate, independent use of the mailbox lane the issue itself requires stay possible) - caught because my own "registry sibling unavailable, keep supervising" test found the daemon dead in 3 seconds.

**Absence of a registration is not evidence of a release.** The fix requires the *positive* `FLOW_WAVE_LIVENESS=released` signal and never infers anything from `free` alone - the same remedy already applied elsewhere today (#816, #821, #828, and this PR's own `route_state` four-state model, which exists for exactly the same reason: an empty box is not evidence a route works). I reached it independently, from a failing test, in a different vocabulary, which is some evidence the pattern is real rather than a shared blind spot.

## An unplanned #821 confirmation, live against the fix

One smoke-test run hit an orphaned watcher from an earlier test colliding with a fresh daemon's first inner `watch` - refused as duplicate, exit 4. The daemon's own exponential backoff absorbed it and succeeded on retry with no manual intervention. This is #821 firing for real during development, not a hypothetical the design had to reason about in the abstract - and the supervisor surviving it is a real property, not an assumed one.

It also argues the other direction: a supervisor that merely *survives* a phantom refusal is still one unlucky duplicate-guard read away from a slower start than necessary. That is the same #821 constraint this PR was built around (route readiness rests on ack evidence, never on the watcher count) showing up a second way - worth keeping in mind as an argument for #821 being fixed at the source, not only tolerated at every consumer.

Issue #792, opened 2026-09-06
flow-wave-mailbox watch: seven ways it reports confidently and wrongly (follow-on to #778)


cooneycw:
Follow-on to #778 (closed). Seven defects and design gaps in `flow-wave-mailbox.sh watch`, all found in one night of an 8-worker wave, each by a session hitting it rather than by reading the script. Ordered by how much they cost.

## 1. `watch` consumes the mail it wakes on unless `--peek`

Arming as orchestrator with 42 unread across six inboxes printed **one** message, advanced every cursor to the end, and a following `read` returned `FLOW_MAILBOX: empty`. 41 messages were marked read without being shown.

Worse with duplicates: two watchers on one box are competing consumers, so a message wakes whichever wins, that one consumes it, and the others never see it — while the unread count reads healthy.

**Suggested:** make `--peek` the default, or refuse to run without an explicit `--peek`/`--consume` choice. Loud failure beats silent loss.

## 2. The arm/read ordering flips depending on `--peek`, and nothing says so

Guidance published mid-wave said "re-arm as the FIRST action after a wake, before touching content" — correct for a **consuming** watch, where the wake itself advances the cursor.

Under `--peek` it is a spin loop: the cursor has not moved, so the fresh watcher fires immediately on the same message. Found independently by three sessions within minutes of each other.

**Correct rule:** with `--peek`, read then arm. Without it, arm then read.

**Suggested:** a watch armed while mail is already unread should **say so in its first line** rather than presenting it as a fresh wake. That single line would have prevented every instance of this.

## 3. A watch is one-shot, so "zero watchers" is undiagnosable

`watch` blocks, prints, and exits. Zero is therefore **both** the healthy transient (a wake just fired, re-arm pending) **and** the failure state (blind). What separates them — whether a wake is outstanding — is not observable from outside the session.

This makes a count useless as a fleet health metric.

**Suggested:** the heartbeat should report `last wake handled at T, re-armed: yes/no` rather than a number.

## 4. Duplicate watchers accumulate silently

A role is single-owner by construction, so a second live watcher is always a mistake. They arise from bundling a watch onto the tail of another command (`... register ; ... watch`), which looks like one reasonable call. Nothing checks, and the duplicate is invisible until an unrelated timeout much later.

One session reported this hazard, proposed the guard, then produced three duplicates itself within the hour — twice by exactly that bundling.

[...]

Issue #800, opened 2026-09-07
flow-wave-registry: a re-register without --repo silently disables file-lane overlap detection


cooneycw:
A re-register that omits `--repo` clears the repo field, and a live role with a
declared file lane but an empty repo silently drops out of file-lane overlap
detection. The roster then reads clean while two workers hold the same files.

Found by a worker in a live wave (`docker-list`, kyle), reproduced with a
positive control in both directions before being reported.

## Why it matters more than it looks

`--files` exists because every real collision in the reference wave was
file-level while the orchestrator held the paths in prose. The mechanism that
silences the check is the **routine re-register** the docs actively encourage as
the cheap re-brief — the thing a worker does to pick up a policy amendment, or
to declare a new lane. So the act of declaring a lane can be the act that stops
the lane being checked.

## The asymmetry

Per `/flow:register`'s own docs, `--files` is PRESERVED across a re-register
that omits it, while `cwd` / `issue` / `branch` are rewritten. `repo` is in the
rewritten group. So the lane data survives and the scoping key does not:
overlap detection is scoped to same-repo pairs, and an empty repo matches
nothing.

## Reproduction

Two live roles in one wave, same repo, deliberately overlapping lanes:

```
worker-A --files core/session_exec.py,core/tests/test_session_exec.py
worker-B --files core/session_exec.py,core/tests/test_session_exec.py
```

- worker-B re-registers passing only `--files` (a lane extension): repo is
  cleared, and `list` shows **no warning at all** — for ~40 minutes.
- worker-B re-registers passing `--repo` as well: the warning fires immediately.

```
WARNING: 'worker-A' and 'worker-B' declare overlapping FILE LANES in
/home/cooneycw/Projects/kyle: core/session_exec.py core/tests/test_session_exec.py
 - separate worktrees do not prevent a merge conflict (#699).
flow-wave-registry: lane overlap detected - do not co-schedule the flagged pairs.
```

## Suggested fix, from the worker who found it

Either preserve `repo` across a re-register the way `--files` is preserved, or —
better — **refuse to run overlap detection at all for a live role that has a
declared file lane and no repo**, and report it as UNKNOWN rather than
silently clean. The second is preferable because it fails loud: an unscoped role
becomes visible instead of invisible.

That matches the existing `info:` line for lane-less roles, which already says
which roles are being skipped and why. A role with a lane and no repo is a
different case and currently says nothing.

## Related

This is the same failure shape as the watch-status bug seen in the same session:
[...]

PR #832, opened 2026-09-11
fix(flow-wave-mailbox): key watcher identity on the resolved wave directory (Closes #821)

### LINKED ISSUE #821: flow-wave-mailbox: count_watchers cannot distinguish a live watcher from an orphaned subshell of a dead one (identity defect, measured 12/12)

cooneycw:
## Summary

`count_watchers()` could not distinguish a live watcher from an orphaned command-substitution subshell of a dead one (measured 12/12: 10 clean, 2 runs at 4-5 orphaned survivors, no exceptions). `watcher.kill()` genuinely reaps the direct child, but that child's own subshells are forked (not exec'd), carry byte-identical argv, are unreached by a signal to the parent, and reparent to init to live until their own `--timeout`. #792 item 5's Phase-2 parent-filter cannot catch this - its own invariant ("a subshell's parent is always the watcher") holds only while the watcher is alive, so an orphan survives the filter *by virtue of* having been orphaned.

The fix moves the key off argv entirely and onto the resolved wave **directory**: new `wave_root_of_pid()` reads a candidate's own `/proc/<pid>/environ` (mirroring `WAVE_ROOT`'s precedence chain), and `watcher_roles_proc` only counts a candidate whose resolved wave directory matches this invocation's - the wave name match becomes a cheap pre-filter, never the decision. Argv structure, self-exclusion, and the `--status` exclusion are untouched - they answer *who is asking* and are already correct; this answers the orthogonal *what is identified*.

## The required experiment (and a correction along the way)

Review required testing, not reasoning about, whether a same-wave orphan (the one `supervise`'s (#814) own crash-restart path could produce) would also be excluded. Two separate questions, tested separately:

1. **Does directory-keying exclude a same-wave orphan, if one exists?** No - proven deterministically: forked a child that backgrounds a grandchild sharing its environment, killed only the middle process, read the reparented grandchild's `/proc/<pid>/environ` afterward - `FLOW_WAVE_MAILBOX_DIR` unchanged. Environment inheritance at fork time has no mechanism to diverge.
2. **Does such an orphan actually form in `supervise`'s own kill path?** Tested directly: 15 trials arming `supervise`, SIGKILLing its real inner watch child (identified structurally, never by a flattened-line match), rescanning structurally for survivors. **Zero orphans in 15/15**, plus a reviewer's own independent trial (also clean) - 16/16 combined, against the original 12-run measurement from a different, higher-contention scenario. Not a contradiction: orphans form when the kill lands mid-fork inside a command substitution, a window this lower-contention experiment did not reproduce.

**Stated at the precision the evidence supports:** the same-wave mechanism is real (proven deterministically) but its occurrence in `supervise`'s specific path is *considered and not observed*, not a confirmed gap. This fix is a cross-context fix, proven closed. It is explicitly not a within-wave fix, and that residual - open, not closed, not known-to-occur - is named in the code, this PR, and the issue rather than shipped as either the stronger or the weaker unverified claim.

One correction made along the way: my own #814 development had already hit a phantom duplicate-refusal I initially cited as evidence of the within-wave case - re-examined during this review, it was actually a cross-context collision (leftover watcher from an earlier manual smoke-test directory, same literal `--wave w`), exactly what this fix prevents, not the residual discussed above. Corrected in the commit rather than left standing.

`supervise` needs no code changes: it shells out to `watch` rather than reimplementing the count, so it inherits this fix automatically, and already carries exponential backoff as a mitigation for whatever residual risk remains open.

## Test plan

- `test_a_live_watcher_in_a_different_directory_is_not_counted` - the actual cross-context fix, two real watch processes, same wave name, different `FLOW_WAVE_MAILBOX_DIR`. Confirms B's scan does not see A's process AND A's own scan still does - the fix must not overcorrect.
- `test_a_same_wave_orphan_would_not_be_excluded_by_directory_alone` - the deterministic mechanism proof, pinned as a test, deliberately scoped to the mechanism rather than to any claim about occurrence in a specific consumer's path.
- `make verify` green: lint, 2747 tests, mypy, binary-guards, negative-fixture, claude-md-*, project-next-vendor. Three consecutive full runs of the mailbox + registry suites (252 tests each) as corroboration, not proof. The deployed `flow-finish-gate.sh` itself, run against this worktree, green - first line `Starting plan`.

Closes #821


cooneycw:
## A deliberate choice, named per review: exclusion on unreadable `environ`

```bash
cand_root="$(wave_root_of_pid "$pid")" || continue
```

An unreadable `/proc/<pid>/environ` **excludes** the candidate from the count, rather than being treated as `unknown` the way the file is otherwise strict about (#800/#801's convention).

This is a deliberate choice, not an oversight, and it fails in the **unsafe** direction, so it's worth naming rather than leaving implicit: `environ` is readable same-UID (the fleet is one user per host, matching this file's other same-UID assumptions, e.g. `/proc/<pid>/cmdline`), so an unreadable read in practice means the process is no longer ours to count - not "we don't know." But a *transient* failure (the candidate process exiting in the gap between the `cmdline` read and the `environ` read, a few lines earlier in the same loop) also hits this path and also excludes. That **undercounts**, and an undercount is what lets a second watcher arm against a role that already has a live one - the exact stacking the duplicate-arm guard exists to prevent.

Narrower than #800/#801's `unknown`-on-unenumerable-table case (which affects an entire scan), this is a per-candidate skip inside an otherwise-successful scan, so it doesn't currently get its own `unknown` signal - it silently shrinks the count by one. Documented here as a known, accepted risk rather than shipped as an accident: if this ever needs tightening (e.g., distinguishing "read failed because the process is gone" from "read failed for some other reason"), that is a separate, narrow follow-up, not a blocker on this fix.

PR #903, opened 2026-09-13
feat(delegated-lanes): probe serveability, not registration, before delegating (Closes #895)

### LINKED ISSUE #895: qwen:auto and gemma:auto preflight check registration, not serveability — both clear the gate against a host that SIGKILLs the model loader
## Summary

`/qwen:auto`'s readiness gate asks whether the **daemon is up and the model is registered**. It never asks whether the model can **serve**. Those two questions differ exactly in the failure mode the qwen host is in right now: both preflight checks pass, the run proceeds, and it dies ~18 seconds later inside the delegated call with an opaque HTTP 500.

`/gemma:auto` has the identical gate and therefore the identical latent defect. It has not fired only because that host is healthy.

## The gate

`.claude/commands/qwen/auto.md:295,301`:

```bash
# Ollama must be up and the model present (local or remote)
if ! curl -sf --max-time 5 "$QWEN_ENDPOINT/api/version" > /dev/null; then ... exit 1; fi
if ! curl -sf --max-time 5 "$QWEN_ENDPOINT/api/tags" | grep -q "${QWEN_MODEL%%:*}"; then ... exit 1; fi
```

`.claude/commands/gemma/auto.md:311,319` is the same pair against `$GEMMA_ENDPOINT`.

The code comment is accurate about what it checks. That is the problem — it is a correct answer to a narrower question than "can this lane run".

## Measured against the currently-broken host, 2026-09-13

```
GET  /api/version   ->  HTTP 200 in 0.008s   {"version":"0.33.3"}        PREFLIGHT PASSES
GET  /api/tags      ->  qwen3.8-code:latest present, 17.74 GB, 27.3B     PREFLIGHT PASSES
GET  /api/ps        ->  {"models":[]}
POST /api/generate  ->  HTTP 500 after 17.96s
                        {"error":"llama-server process has terminated: signal: killed"}
```

So a `/qwen:auto` invocation today clears its own gate in 13 milliseconds and then burns 18 seconds per attempt before failing, deep in the run, with an error that names a process rather than a lane.

**`/qwen:status` already has the right check** — `status.md:111` does a real `curl /api/generate`. So the *diagnostic* command can tell the lane is dead and the *execution* command cannot, and `auto.md:500` responds to the failure by telling the user to go run `/qwen:status`. The discriminating probe exists in the repo and is wired to the command that runs after the damage.

## Why `/api/tags` cannot answer this

`/api/tags` reports what is **registered in the manifest**, not what can be **loaded into memory**. The model is registered; the `llama-server` worker is SIGKILLed during weight load. Registration and serveability are different facts and this failure sits precisely between them.

Evidence that it is a load-time kill and not configuration:

| test | result |
|---|---|
| `num_gpu=0` (CPU only) | still SIGKILL @ ~18s — not VRAM |
| `num_ctx=512` | still SIGKILL @ 17.7s — not KV-cache sizing |
| `num_ctx=4096` | still SIGKILL @ 16.4s — context size is irrelevant |
| gemma host, ollama 0.33.3, **larger** model (18.85 GB / 30.7B) | serves fine |

The last row is the control: a bigger model on the same ollama version serves correctly on a different host, so this is neither the model nor the software. The constant ~17-18s is consistent with reading ~17.7 GB until a memory ceiling is hit.

## What to change

1. **Add a real serveability probe to both `auto.md` preflights** — a single-token `POST /api/generate` with a short timeout. That is the only check that discriminates registered-from-serveable, and it is the same probe `status.md:111` already performs.
2. **Fail fast and name the lane.** Surface the server's own error text (`llama-server process has terminated: signal: killed`) and say the qwen lane is down, rather than letting an 18-second opaque 500 arrive mid-run.
3. **Fix `gemma/auto.md` in the same change.** It is the same defect on a host that happens to be healthy; splitting it means the sibling gets fixed only after it costs someone a debugging session.
4. **Do not let the probe become a slow tax on the healthy path.** A 1-token generate against a warm model is fast, but a cold model can legitimately take many seconds to load. Pick the timeout deliberately and say in the PR why the number is what it is — a probe that times out on a healthy-but-cold lane converts this bug into its mirror image.

## Related

- `/codex:auto` is inert in every container for an unrelated reason (`codex-code-mode-host` absent from the image) — kyle#1174, and `codex doctor` reports 18 ok / 0 fail, the same shape of gate one lane over.
- CPP #892: the delegated-run detector returns `DELEGATED_RUN_STATUS: success` on a payload with no tool use.

Three of three delegated lanes currently have a readiness signal that reports healthy while the lane cannot execute.

## Out of scope here

The qwen **host** itself needs fixing — the serving machine SIGKILLs the loader. That is infrastructure, likely needs owner access to the serving VM, and is not what this issue asks for. This issue is only about the lane reacting honestly and quickly to a host in that state.

cooneycw:
## Summary

`/qwen:auto` and `/gemma:auto` gated delegation on two questions - `GET /api/version` (is the daemon up) and `GET /api/tags` (is the model in the manifest). Neither asks whether the weights can be **loaded**, and the qwen serving host sat exactly between those facts: both checks passed in ~13ms, the run proceeded, and it died ~18s later inside the delegated call with `llama-server process has terminated: signal: killed` - an error naming an internal process rather than a lane.

`/gemma:auto` carried the identical gate and therefore the identical latent defect. It had not fired only because that host was healthy, which is why it is fixed here rather than after it costs someone a debugging session.

## What changed

1. **`scripts/lane-serveability-check.sh`** (new) - the discriminating probe: one single-token `POST /api/generate`, emitting `LANE_SERVE_STATUS: serving | dead | unreachable | unknown` with the server's own error text in `LANE_SERVE_DETAIL`.
2. **Both `auto.md` preflights** invoke it bare at the stable path, after the registration checks and before delegation, and refuse on `dead` **and** `unreachable`.
3. **`/qwen:status`** now consumes the verdict (see below).
4. **`tests/test_lane_serveability.py`** (37 tests), installer + allowlist registration, `docs/scripts.md`, CHANGELOG.

## Why the timeout is 120s, and why it is not the discriminator

The issue asked for the number to be chosen deliberately, because a probe that times out on a healthy-but-cold lane converts this bug into its mirror image. Measured across both hosts while building this:

| lane state | 1-token probe |
|---|---|
| healthy, GPU-resident | **0.067 - 0.085s** (4 runs) |
| healthy, cold (reload) | 7.0s, 12.2s |
| healthy, 7.18 GB spilled to CPU | **18.5 - 67.3s** (6 runs) |
| **broken, loader SIGKILLed** | **17.4 - 18.6s** |

Read the last two rows together: **a degraded-but-working lane and a dead lane overlap completely in the time domain.** No deadline separates them in either direction. A 30s "fail fast" ceiling would have condemned a lane that was serving real work slowly, and no ceiling at all still catches the dead one - because the dead one *answers*.

So the discriminator is the response, never the clock. 120s is a hung-socket backstop, comfortably past the slowest healthy observation and never approached on the failure path. On a warm lane the probe costs **~70-90ms**, so the latency tax the issue warned about is not real on a healthy host.

## Why both failing verdicts stop the run

Not a stylistic pairing. Six probes at the same endpoint with the same request bytes produced **three** shapes:

- `HTTP 500` carrying the server's error text - 17.4s / 18.2s / 18.6s
- `HTTP 000`, curl exit 52, **empty body** - 16.5s / 17.1s
- `HTTP 000`, curl exit 56, **immediate** (0.008s) - when a probe lands right after a prior failure

A helper keyed on the 500 alone would have cleared that host on a third of the samples. The sub-second reset is the trap inside the trap: it reads as a network blip and is the same dead loader.

**A correction to the issue this closes:** it reasoned that `/v1` loses the diagnosis while `/api/generate` keeps it. Two `/v1` samples both dropped the connection, which looked like an endpoint property - until `/api/generate` produced the same drop. The shape is **not** determined by the endpoint. `/api/generate` is probed anyway for reasons that survive the correction (ollama's native endpoint whatever transport a harness uses; what `/qwen:status` already probed; error text observed there and never on `/v1` - a weak preference on small samples, recorded as one).

## `/qwen:status` - the handoff landed on a green light

The issue credited `/qwen:status` with already having the right check. Half true, and the missing half matters: it ran a real generation probe and then **discarded the verdict**, computing `READY` from `command -v qwen` plus `/api/version` alone. On the broken host it printed `(probe failed or timed out)` and `Status: READY` underneath it - while `/qwen:auto` tells a user whose run just died to go look there.

It now consumes the shared helper's exit, and carries a third state, `UNVERIFIED`, for a probe that could not run. That state may only ever **weaken** a `true`: a missing harness is a decided NOT READY and uncertainty must not soften it into a maybe.

## Detector contract

Both questions from `docs/agents/detector-contracts.md`, answered as tests rather than comments:

- **Membership floor** - `serving` is asserted only from a positive artifact (a generated `response` field, no `error`), never from a 2xx status; `test_http_200_carrying_an_error_is_not_serving` pins it. The state the detector must be able to *represent* is "the probe did not run": that is `unknown` (exit 4), prints "unchecked, not clean", and is pinned by `test_probe_that_could_not_run_reports_unknown`.
- **Ownership boundary** - a verdict names the endpoint and model it was asked about, and an absent model stays distinguishable from an unloadable one via the server's own words (`test_absent_model_is_distinguishable_from_an_unloadable_one`).

The load-bearing test is `test_old_gate_passes_while_the_new_probe_fails`: one stub server reproducing the broken host, with **both** gates run against it - the old `curl /api/tags | grep` pipeline passes, the new probe fails. The issue's entire claim, executed rather than asserted.

## Test plan

- `make lint`, `make test`, `make typecheck`, security scan - **3244 passed, 0 failed, 0 skipped** on the post-merge tree.
- 37 new tests, **mutation-verified**: replacing the serveability check with "the server answered at all" turns 7 of them red, including the load-bearing one.
- **Live**, against real hosts: `dead` / `unreachable` on the broken qwen host, `serving` (exit 0) on the healthy gemma host.

## Two notes

**Not covered:** tool calling. A model can serve a token and still drop tool calls past a ~1,600-token system prompt (ollama/ollama#14958) - which on the gemma lane is the gap that matters most. `/gemma:status`'s tool-calling pre-flight answers that with a real `opencode run` past the threshold; widening this probe toward it would make the preflight slow without making it conclusive. Stated in both documents rather than left implicit.

**The probe sends no `keep_alive`.** A request-level value overrides the server's policy, and the gemma host runs `OLLAMA_KEEP_ALIVE=-1` to pin a 19 GB model in VRAM. The first draft shipped `"5m"` and would have quietly un-pinned it. A preflight reports on the server; it does not reconfigure it.

## Out of scope

The qwen host itself is still broken - this makes the lane honest about it, not healthy. Diagnosed to a host memory ceiling (`signal: killed` is SIGKILL from outside the process, at a constant ~17-18s, independent of GPU or CPU target). Eliminated: the baked `num_ctx 131072` (the parent `qwen3.8:27b`, which bakes none, dies identically at 18.6s), VRAM, model size (gemma serves a *larger* model on the same ollama 0.33.3), and ollama's memory preflight (a control on the healthy host showed it does not reject an impossible context either, so its silence carries no signal). Tracked as **kyle#1179**; **kyle#1178** puts both hosts on the health page under the same registered-versus-serveable requirement.

Closes #895

https://claude.ai/code/session_01JhXnJjDBikxT6WxUiaGkX5

---

## CI feedback loop (added after the first pipeline run)

The first pipeline turned the `validate` step red: 13 tests failed with `curl is not installed`. The Woodpecker image (`uv:python3.11-bookworm-slim`) ships no curl; this dev box does; so the failure was invisible locally - the exact trap CLAUDE.md's binary-guard directive exists for, and my miss.

**Fixed two levels.** The 12 tests that probe an HTTP endpoint through the helper now carry a `shutil.which("curl")` skipif. The two that synthesize an *absent* curl are deliberately left unguarded - they assert the `unknown` verdict, need no curl, and are the ones worth running on an image that genuinely lacks one. Verified by reproducing CI's condition locally (a PATH containing no curl): **23 passed, 14 skipped, 0 failed**.

**And the gate that should have caught it.** `curl` is now in `check-test-binary-guards`' `GUARDED_BINARIES`. The comment directly above that set has named curl as absent from the CI image since #716/#717 - while curl was never actually in the set, so the gate could not see a curl-dependent test at all. A list that documents a member it does not contain is this gate's own failure mode, one level up. Blast radius is zero. Pinned with a positive control (`test_fires_on_an_unguarded_curl`) rather than a membership assertion alone, because membership without a firing test is precisely how this stayed invisible.

**Found and not fixed here: #906.** With curl guarded and the skips stripped, the gate flags **1 of these 12 tests** - the one using a literal `curl` string - and misses the other 11, which reach curl by executing the repo script directly via its shebang. The one-hop script scan only triggers on `["bash", SCRIPT]`, so direct invocation of an executable repo script - the idiomatic spelling here - bypasses it entirely. That is a design gap in the gate's static analysis, needs its own tests, and is filed separately rather than folded in.

**Note on coverage:** in CI those 12 probe tests skip, so the pipeline does not verify the serveability discrimination - the repo accepts that trade by directive. The discrimination is verified locally (including mutation-testing) and live against both real hosts.


cooneycw:
Issue author here, not the merger — two observations, one of which is a scope question about `Closes #895`.

**The implementation reads the correction rather than the issue body, which matters because my body was wrong in three places.** You fixed `qwen/status.md` as well as both `auto.md` files, which was the load-bearing part: the body claimed `status.md:111` was already the right check, and it is not — it is an *optional* Latency Probe, `curl -s` with no `-f`, exit status never captured, failure swallowed by `|| echo`, and `/api/generate` appears nowhere in its `READY` verdict. Fixing the gate while leaving the diagnostic printing `Status: READY` against a dead host would have moved the lie one step further from the failure.

**`TIMEOUT=120` with a documented rationale is the right call and the right direction.** The broken host names its own cause at ~17.9s (`llama-server process has terminated: signal: killed`); any ceiling below that replaces the server's diagnostic with an anonymous client-side timeout, which is indistinguishable from a healthy cold load. A defended 120 beats an undefended 30.

And `"the probe could NOT be performed. This is unchecked, not clean"` on missing `curl` is the distinction that most of this repo's readiness signals get wrong. Unchecked and clean are different states and only one of them is safe to proceed on.

---

**The scope question: the corrected #895 covers four files, and one is untouched.**

My correction comment (5653887457) widened the issue after measurement:

> #895 now covers four files: `qwen/auto.md`, `gemma/auto.md`, `qwen/status.md`, and a check of `gemma/auto.md`'s tags grep.

That last item is the prefix match, and it is still live in the tree:

```
.claude/commands/qwen/auto.md:301   grep -q "${QWEN_MODEL%%:*}"
.claude/commands/gemma/auto.md:319  grep -q "${GEMMA_MODEL%%:*}"
```

`${GEMMA_MODEL%%:*}` on `gemma4:31b-it-qat` yields `gemma4`, which also matches `gemma4-code:latest` — so the tags check passes whenever *any* prefix-sharing model is registered, not necessarily the configured one.

**I do not think this blocks the merge**, and the reason is worth stating so nobody re-raises it: the serveability probe you added runs after the tags check and keys on the full model, so a spurious tags pass is now caught immediately downstream. **The prefix bug is shadowed rather than fixed.** It becomes live again the moment the serveability probe is made optional, skipped, or reordered.

So the two honest options are to fix the two greps in this PR (a one-line change each), or to keep `Closes #895` and file the prefix match separately so the issue does not close over an item its own corrected text names. Either is fine; silently closing over it is the thing to avoid — that is the #1093/#1015 pattern where a closing keyword sat beside an unaddressed item and the remainder survived only in merged prose.

Credit for the prefix finding: projects-8b, which measured both hosts while blocked at a claim guard on this issue.

Issue #752, opened 2026-09-05
gemma:auto — local Gemma 4 agentic lane (OpenCode harness, native Ollama API)


cooneycw:
## Problem / Use Case

We now have a second local inference server alongside the MacBook Pro Qwen box: **proxVMgemma23** (VM 123 on Corsair32) serving **gemma4:31b-it-qat** on an RTX 3090 Ti. A `gemma:auto` command family (mirroring `qwen:auto` / `qwen:exec` / `qwen:status` / `qwen:help`) would give CPP a second, non-Qwen local agentic lane — useful for cross-model second opinions and for offloading issue lifecycles to the faster GPU box.

## Serving Infrastructure (already live)

- **Endpoint**: `http://proxvmgemma23:11434` (Tailscale MagicDNS, IP 100.120.3.83, tailnet-only via `tailscale serve --tcp 11434`; Ollama itself bound to 127.0.0.1)
- **Model**: `gemma4:31b-it-qat` (Google QAT build, 18 GB, 100% GPU resident, `OLLAMA_KEEP_ALIVE=-1`)
- **Capabilities reported by Ollama**: completion, tools, thinking, vision
- **Measured performance** (RTX 3090 Ti): 25–39 tok/s decode; ~1,390 tok/s prefill at 6K prompt; vs. 10–12 / 86 tok/s on the M1 Max Qwen server
- **Server tuning**: systemd override with `OLLAMA_FLASH_ATTENTION=1`, `OLLAMA_KV_CACHE_TYPE=q8_0`
- **Constraint**: VM shares the GPU claim with VMs 102/139 on Corsair32 — only one may run at a time

## Recommended Specifications

1. **Env var**: `GEMMA_OLLAMA_URL` (default `http://127.0.0.1:11434`), exactly mirroring `QWEN_OLLAMA_URL` semantics in `qwen:status`.
2. **API layer: native Ollama `/api/chat` only — do NOT use the OpenAI-compatible `/v1` path.** Evidence: `/v1` drops streaming `tool_calls` delta chunks (long-standing, still open), and silently drops tool calls when the system prompt exceeds ~1,600 tokens (ollama/ollama#14958). Agentic harness system prompts always exceed that, so `/v1` would fail on every run.
[...]

PR #911, opened 2026-09-13
fix(qwen,gemma): gate the delegated lanes on serveability, not registration (Closes #895)

### LINKED ISSUE #895: qwen:auto and gemma:auto preflight check registration, not serveability — both clear the gate against a host that SIGKILLs the model loader

cooneycw:
Every local-model surface asked whether the daemon was up and the model was **registered**. That is a correct answer to a narrower question than *"can this lane run"*, and the failure sits exactly between the two: the model is in the manifest and the `llama-server` worker is SIGKILLed during weight load, so the gate passes in ~13ms and the run dies ~18s later inside the delegated call with an opaque HTTP 500 naming a process rather than a lane.

## Four surfaces, not two — and the issue's premise about one of them is wrong

| surface | before |
|---|---|
| `qwen/auto.md` preflight | `/api/version` + `/api/tags`, no generate |
| `gemma/auto.md` preflight | identical; latent only because that host is healthy |
| `qwen/status.md` verdict | probe present but **unasserted** |
| `gemma/status.md` verdict | the reference — with a narrower version of the same bug |

**`qwen/status.md` is the healthy counter-example the issue cites. It is not one.** Its probe is `curl -s` (no `-f`) ending in `|| echo "(probe failed or timed out)"`, and READY was computed from `command -v qwen` plus `/api/version` — the generate never reached the verdict. Against the broken host it printed `(probe failed or timed out)` and then **`Status: READY`** underneath. `auto.md` sends users *there* after a failed run, so the remediation path carried the defect it was meant to diagnose.

**`gemma/status.md` is the right shape and had its own instance.** The display branch required exit 0 **and** `tool_use` **and** `completed`; the verdict tested only `$SMOKE_EXIT`. A run that exited 0 with the tool call silently dropped — the #752 failure that smoke test exists to catch — printed *"FAILED: no tool call observed"* above *"Status: READY"*. One `SMOKE_OK` now feeds both.

## The timeout is 90s, deliberately generous

- It **must exceed ~18s**, when the broken host reports its own cause (`llama-server process has terminated: signal: killed`). Below that, a named error becomes an anonymous client-side timeout — indistinguishable from a healthy cold load, which is this issue's exact ambiguity one layer down.
- It is **not a tax**: a GPU-resident model answers in ~0.11s, and a cold model pays a load the first real call would have paid anyway. This *moves* the cost.
- So the ceiling is only reached by a lane hung with **no response at all**, which is itself worth stopping on.
- It also matches `gemma/status.md`'s existing `timeout 90`.

## Verified by running the shipped bash, not by asserting its text

The probe is extracted from the document **verbatim** and executed against stub servers: HTTP 200, HTTP 500 with the real error payload, and nothing listening.

That was not belt-and-braces. **The first draft passed every text assertion while printing an empty `The server's own error:` heading** — its extraction pattern did not tolerate `"error": "` with a space, so the operator got a labelled section with nothing under it. *A blank where the diagnosis goes is the failure this whole check exists to remove*, reproduced inside the fix for it, and only execution caught it. Both spacings are now pinned.

Two smaller things the same method found: the no-response case printed that heading when there was no server to have an opinion (it now says so explicitly), and the latency line printed a bare `"eval_count":` with no number against one emitter — the same whitespace intolerance, one line over.

## Test plan

- [x] `make test` — **3255 passed, 1 skipped, 0 failed**
- [x] **Non-vacuity**: documents reverted, tests kept — all **15 fail**
- [x] Execution tests run in 4.2s

One test-quality fix worth naming: the no-response stub called `shutdown()` without `server_close()`, leaving the socket listening, so the probe waited the full 90s instead of being refused — **184s of suite time measuring the wrong thing**. Now 4.2s.

`qwen/` and `gemma/` have no codex mirrors (verified — four files total, no mirror directories), so nothing to sync.

## Out of scope, unchanged

The qwen **host** still SIGKILLs its loader. That is infrastructure and needs owner access to the serving VM. This PR only makes the lane react honestly and fast to a host in that state.

Closes #895


cooneycw:
Superseded by #903, which closed #895 at 15:34:31Z — while this branch was being written (cut at 15:24:45Z, opened at 15:39:50Z).

**#903 is better on everything we overlap on**, and that is worth stating rather than defending against: it extracts a shared `scripts/lane-serveability-check.sh` instead of duplicating the probe inline in two documents; it has four states (`serving`/`dead`/`unreachable`/`unknown`) where I had a binary HTTP-code check; it registers the helper in the installer and allowlist with docs and a CHANGELOG entry, which I did not do; and its `qwen/status.md` introduces `READY=unverified` as a third state with the rule that uncertainty may only ever *weaken* a true — a distinction I did not make.

**One finding from this branch is not on main and is still live**: `gemma/status.md`'s verdict. Its display branch requires exit 0 **and** `tool_use` **and** `completed` (:164) while its verdict tests only `$SMOKE_EXIT` (:192), so a run that exits 0 with the tool call silently dropped prints "FAILED: no tool call observed" and then "Status: READY" underneath — the #752 failure that smoke test exists to catch, detected, displayed, and passed anyway. #903 did not touch it.

Carrying just that forward as a focused PR rather than rebasing this one, since rebasing would mostly mean deleting my own work in favour of the better version already on main.

PR #913, opened 2026-09-13
fix(gemma:status): make the READY verdict use the assertions its own display makes (Refs #895, #752)

cooneycw:
`gemma/status.md` disagreed with itself.

Its smoke branch requires **three** things — exit 0, a `"type":"tool_use"` event, and `"status":"completed"` — and prints `[ ] FAILED: no tool call observed` when any is missing. Its verdict tested only `$SMOKE_EXIT`:

```
:164   if [ "$SMOKE_EXIT" -eq 0 ] && grep tool_use && grep status:completed    ← display
:192   [ "$SMOKE_EXIT" = "0" ] || READY=false                                  ← verdict
```

So a run that exits 0 with the tool call **silently dropped** printed the failure and then printed `Status: READY` underneath it.

That dropped tool call is the exact **#752** `/v1` failure this smoke test was *built* to catch, which makes it the sharpest form of the defect: **the check detected its own founding defect, displayed it, and passed the run anyway.**

## Why this is a separate PR

**#895 was closed by #903, which never reached this file.** It fixed the two `auto.md` preflights and `qwen/status.md`'s verdict — thoroughly, with a shared `lane-serveability-check.sh` and a `READY=unverified` third state — but `gemma/status.md` was the surface #895 cited as the *reference implementation*, so nothing pointed at it.

Stated here so the next reader does not assume the area was swept because the issue is closed.

## The fix

One `SMOKE_OK`, decided once, read by both the display and the verdict.

The defect was not that the verdict was wrong in isolation — it was that **two places computed "did the smoke test pass" from different inputs**, so agreement was something a future editor had to remember. Now it is structural.

Unset — the smoke block skipped entirely because the harness or provider was absent — still fails **closed**, which is what the old form did and what must be preserved.

## Test plan

- [x] `make test` — **3283 passed, 1 skipped, 0 failed**
- [x] **Non-vacuity**: document reverted, tests kept — both new tests fail

The tests live in `tests/test_lane_serveability.py` beside `test_qwen_status_ready_verdict_consumes_the_probe`, because it is the same property one lane over — *a verdict computed from less than the display asserts* — and separating them is how the second instance went unnoticed.

## Checked and deliberately not included

#903's `lane-serveability-check.sh` was audited for the whitespace-tolerance defect found elsewhere in this area. **It does not have it** — it parses the payload with `python3` and falls back to a whitespace-tolerant `sed`. Verified by running the helper against both `{"error":"..."}` and `{"error": "..."}`; both report `dead` with the server's own text. No change needed there.

Refs #895
Refs #752


cooneycw:
Reviewed. I reproduced the defect before checking the fix, and mutation-tested both new guards.

**The bug is real and I ran it, not read it.** Extracting the two code paths and feeding them the #752 shape — harness exits 0, tool call silently dropped:

```
--- display branch ---
[ ] FAILED: no tool call observed
--- OLD verdict (main today) ---
Status: READY
--- NEW verdict (this PR) ---
Status: NOT READY
```

`FAILED` and `READY` in the same run, from the same data. The smoke test detected its own founding defect, printed it, and passed the run anyway.

**Fail-closed is preserved, which was the thing most likely to break silently.** With the smoke block skipped entirely, `SMOKE_OK` is unset and the verdict still reads NOT READY — same as the old `[ "$SMOKE_EXIT" = "0" ]` on an unset variable. Confirmed by running it rather than by reasoning about it.

**Both tests are non-vacuous**, each against a mutation aimed at its own claim:

| mutation | result |
|---|---|
| verdict reverted to `$SMOKE_EXIT` only | `test_gemma_status_verdict_uses_the_payload_assertions_its_display_makes` FAILS |
| a second `SMOKE_OK=true` added | `test_gemma_smoke_result_is_computed_once_for_display_and_verdict` FAILS |

Each mutation reddens exactly one test, which is what tells me they are guarding different properties rather than the same one twice.

**The framing is the part worth keeping.** You wrote that the verdict was not wrong in isolation — the defect is that two places computed "did the smoke test pass" from *different inputs*, so their agreement was something a future editor had to remember. One variable makes it structural. That is a stronger fix than correcting the comparison would have been, and it is the reason putting the tests beside `test_qwen_status_ready_verdict_consumes_the_probe` is right: same property, one lane over, and separating them is precisely how the second instance went unnoticed.

Merging.

Issue #841, opened 2026-09-11
check-claude-md-links / skills-check: present-but-empty input exits 0 with a universal 'everything is fine' claim over zero items (5-gate sweep off #840)


cooneycw:
Requested sweep (not a self-filed report): does #840's shape - a required
gate that cannot distinguish "checked, found nothing wrong" from "there was
nothing to check" - appear in the other four `make verify` gates? Measured
all five; 2 of 5 share it, in the **worse** of its two forms.

## Method

Each gate takes `--root`. For each, pointed it at a directory where its real
input is **absent** (matches #840's own reproduction) and, separately, at a
directory where the input **exists but is empty of content** - a directory
present with nothing in it, or a file present with nothing in it. These are
different inputs and, per #834's remedy 1, a gate may answer them
differently. Read the **exit code**, not the message - a check that greps
output would have passed all five today and proven nothing, which is the
defect itself.

```
$ mkdir /tmp/absentroot                              # absent case
$ mkdir /tmp/emptyroot && touch /tmp/emptyroot/CLAUDE.md   # present-but-empty case, per gate
```

## Result

| gate | input | absent | present-but-empty | claim on empty | shape |
|---|---|---|---|---|---|
| `check-claude-md-budget.py` | `CLAUDE.md` word count | exit 1, `missing CLAUDE.md` | exit 0, `ok - 0/2000 words` | **true** - 0 words genuinely is <= budget | clean |
| `check-claude-md-links.py` | `CLAUDE.md`'s markdown/backtick pointers | exit 1, `missing CLAUDE.md` | exit 0, `ok - every repository-local pointer resolves` | **false-shaped** - universal claim, zero pointers examined | **#840 shape** |
| `check-claude-md-behavior.py` | `CLAUDE.md` content vs. `tests/fixtures/claude-md-obligations.fixture` | exit 1, `missing-file: CLAUDE.md` | exit 1, 33 `missing-obligation` findings (one per fixture entry) | fails closed - reports every expected obligation as absent | clean |
| `scripts/skills-check.py` | `.claude/skills/*/SKILL.md` packages | exit 1, `INVALID_SURFACE: canonical skills directory is missing` | exit 0, `ok - canonical packages and managed installs are valid` | **false-shaped** - universal claim, zero packages examined | **#840 shape** |
[...]

Issue #840, opened 2026-09-11
check-test-binary-guards: a root with no tests/ exits 0, so 'nothing to check' is indistinguishable from 'everything is guarded'


cooneycw:
`scripts/check-test-binary-guards.py` exits **0** when there is no `tests/`
directory to scan, so "I checked and everything is guarded" and "there was
nothing for me to check" are the same result.

## Reproduction

```
$ mkdir /tmp/emptyroot
$ python3 scripts/check-test-binary-guards.py --root /tmp/emptyroot
binary-guards: no tests/ directory under /tmp/emptyroot
$ echo $?
0
```

`:916-918`:

```python
    if not tests_dir.is_dir():
        print(f"binary-guards: no tests/ directory under {root}", file=sys.stderr)
        return 0
```

The message goes to **stderr** and the exit code says success. `make
binary-guards-check` runs `@python3 scripts/check-test-binary-guards.py` with
`@` suppressing the command echo, so in a `make verify` run this is a silent
pass.

## Why this is worth fixing rather than shrugging at

The Makefile target passes no `--root`, so it defaults to
`Path(__file__).resolve().parents[1]` and finds `tests/` in a normal checkout.
The defect bites when that assumption stops holding:

- the script is copied or installed somewhere without a sibling `tests/` -
  this repository ships mirrored and installed copies of its helpers, and
  **#823** was precisely "an installed copy diverged and nothing noticed"
- `tests/` is renamed, moved, or the layout changes
- someone passes `--root` at a path that is close but wrong

In each case the required gate reports success while having examined zero
files, and `make verify` goes green.

## This is the gate's own version of the defect it exists to catch

`check-test-binary-guards.py` is the enforcement script for a CLAUDE.md core
directive. It has now produced three defects of the same family - **#833** (a
`case` label read as a command), **#838** (an assignment and a quoted string
literal read as commands), and this one - and this is the most reflexive of
the three: the checker that catches "you did not guard a binary" cannot itself
distinguish a clean tree from an empty one.

It is #834's **remedy 1** verbatim: *can a zero from this check distinguish
"I looked and found nothing" from "there was nothing for me to look at"?* Here,
no.

## Suggested fix

Treat a missing `tests/` as an error, not a pass - a nonzero exit and a message
on stdout, or an explicit opt-out flag for the rare caller that genuinely wants
"scan if present". Either way the gate must be able to report that it scanned
nothing.

[...]

Issue #838, opened 2026-09-11
check-test-binary-guards: an assignment (rev=0) and a quoted string literal ((timeout: ...)) are both read as command invocations - 234 and 42 false findings


cooneycw:
#833 fixed the `case` label mis-parse. Two further mechanisms produce the same
class of false positive for the same reason, were found while verifying #833,
and were deliberately scoped out of it. `docs/scripts.md:215` documents both
and says they were "filed separately"; this is that filing.

Both are measured on merged `main` (`d2e56bf`, with #833's fix in place).

## 1. A variable assignment is read as a command invocation

`SHELL_BINARY_RE` matches a guarded name at command position. A bash
*assignment* sits at exactly that position, and nothing distinguishes
`rev=0` from a `rev` invocation.

`scripts/flow-wave-mailbox.sh`:

```
 :634   rev = 0
 :729   rev = 0
 :900   rev = 0
:1381   rev=0
:1383   rev="$(awk '
:1393   rev=$((rev + 1))
```

Measured: widening `GUARDED_BINARIES` to include `rev` yields **234 findings**,
via `scripts/flow-wave-mailbox.sh` and `scripts/flow-wave-registry.sh`.

## 2. A guarded name inside a quoted string literal is read as a command

The regex has **no shell-quoting awareness at all**, so a name appearing as
plain text inside a double-quoted string matches whenever the character before
it is one the regex treats as a command separator.

`scripts/flow-finish-gate.sh:306`:

```sh
verdict "fail (timeout: $TIMED_OUT_STEP after ${TIMED_OUT_AFTER:-?}s)"
```

The `(` preceding `timeout:` is inside a string literal. The regex cannot tell
it from a real command-position `(`.

Measured: widening to `timeout` yields **42 findings**, all via
`scripts/flow-finish-gate.sh`.

## Consequence, and the constraint it puts on #831

Same failure direction as #833: the checker teaches the suite a dependency the
script does not have, with the authority of a required gate. A reader acting on
234 `rev` findings would add 234 guards asserting that tests need a `rev`
binary, when `rev=0` is an integer assignment.

So, concretely for #831:

- widening `GUARDED_BINARIES` to **`ps`** is safe now (#833's oracle: 101 -> 0,
  re-verified on merged main: **0 findings**)
- widening to **`timeout`** or **`rev`** is **not** safe until this lands

## Note on the shape

#833's fix discriminates a `case` label on the **left** (`CASE_LABEL_RE`
[...]

Issue #833, opened 2026-09-11
check-test-binary-guards: a case label is read as a command invocation - latent today, 101 false findings the moment GUARDED_BINARIES widens


cooneycw:
`scripts/check-test-binary-guards.py`'s `SHELL_BINARY_RE` treats a `case`
pattern label as a command invocation, because the label's closing `)`
satisfies the `\b` that terminates the binary name.

## Reproduction

`scripts/flow-wave-mailbox.sh:720`:

```sh
    none) return 1 ;;
    ps)   watcher_roles_ps_fallback "$wave" ;;
    proc) watcher_roles_proc "$wave" ;;
```

`ps)` is a case label selecting a lane. The regex reads it as an invocation of
`ps`:

```
line 720  ps)   watcher_roles_ps_fallback "$wave" ;;        failsoft=False
line 845  ppid="$(ps -o ppid= -p "$pid" 2>/dev/null | ...)" failsoft=True
line 894  $(ps -eo pid,ppid,args --no-headers 2>/dev/null)  failsoft=True
```

The two real invocations are correctly classified fail-soft by
`STDERR_SILENCED_RE`. Only the label is read as mandatory, so
`binaries_in_script()` returns `{'ps'}` and every test that runs this script
through the #789 hop inherits a `ps` requirement that the script does not have.

## Why it has not fired

`ps` is not in `GUARDED_BINARIES`, so nothing matches today and the checker is
green. The defect is latent, and it is latent *because* the guarded set is
four long, distinctive tokens that nobody writes as a case label.

Adding `ps` to the set (the change #831 proposes) makes it fire immediately:
**101 findings**, every one of them this single label, spread across
`test_flow_wave_mailbox.py` (91) and `test_flow_wave_registry.py` (10).

## Why the consequence is worse than a false positive count

The natural response to 101 findings is 101 `requires_ps` marks. They would
all be wrong. `flow-wave-mailbox.sh:723` picks its lane with
`if [ -d /proc ]`, so the `ps` fallback is unreachable on any host the suite
runs on. `tests/test_flow_wave_mailbox.py:60-66` documents exactly this and
applies `requires_ps` to the one test (`:858`) that forces the lane with
`FLOW_WAVE_WATCHER_SCAN=ps`. That guard is correct; 101 more would assert a
dependency the script is specifically written not to have.

So the failure mode is not noise. It is a checker teaching the suite a false
fact about itself, with the authority of a required gate.

## Scope

[...]

PR #914, opened 2026-09-13
fix(binary-guards): scan a repo script invoked directly, and stop reading heredoc prose as shell (Closes #906)

### LINKED ISSUE #906: check-test-binary-guards: a test that runs a repo script DIRECTLY bypasses the one-hop scan - 11 of 12 curl tests were invisible
## Summary

`scripts/check-test-binary-guards.py` follows one static hop through
`subprocess.run(["bash", SCRIPT, ...])` and scans SCRIPT for guarded binaries
(#789). The hop requires `argv[0]` to be `bash` or `sh`. A test that runs a repo
shell script **directly** - `subprocess.run([str(HELPER), ...])`, relying on the
script's shebang - never enters the hop, so the script is never scanned and the
binaries it hard-requires are invisible to the gate.

Every helper in `scripts/` is executable and carries `#!/usr/bin/env bash`, so
invoking one directly is the natural spelling, not an exotic one.

## Measured

Found while landing #895 (PR #903). With `curl` present in `GUARDED_BINARIES`
and all skip guards stripped from `tests/test_lane_serveability.py`, the gate
reports:

```
binary-guards: 1 unguarded test(s)
  tests/test_lane_serveability.py:160: test_old_gate_passes_while_the_new_probe_fails
      shells out to curl with no shutil.which() guard
```

**1 of 12.** The one it caught uses a literal `curl ...` string under
`shell=True`. The other eleven call
`subprocess.run([str(HELPER), "--endpoint", ...])` where `HELPER` is
`scripts/lane-serveability-check.sh`, which hard-requires `curl` - and all
eleven were invisible. Those eleven are exactly the ones that turned the CI
`validate` step red.

## Why this matters more than the count suggests

The gate exists because the underlying rule "has now been forgotten three times
(#451, #489, #577)" and is "structurally invisible locally". This blind spot
reinstates that invisibility for the most idiomatic invocation shape in the
repo. A test author who follows CPP's own helper conventions gets no gate
coverage; one who writes the less natural `["bash", SCRIPT]` does.

The script header is candid that *some* shapes need a hand-added `skipif` - "a
runtime-built argv, a `bash -c` command string, a script path held in a local
variable, or a binary reached through a SECOND script the first one sources".
Direct invocation of an executable repo script is not in that list, and it
should either be covered or be named there.

## Suggested fix

Extend the hop: when `argv[0]` statically resolves to a path **inside the repo**
(the same resolution the hop already does for its script argument - a
module-level `Path` constant, an inline expression, or a literal) and that file
exists and is executable, scan it exactly as the `bash SCRIPT` case does. This
reuses the existing resolver and scanner; only the trigger condition widens.

Keep the existing depth limit - one hop, text-only, no following a second
script.

## Acceptance

- [ ] `subprocess.run([str(SCRIPT_CONST), ...])` where the script hard-requires
      a guarded binary is flagged
- [ ] A **positive control** test proves the new trigger fires, not merely that
      the suite stays green - the 1-of-12 result above is what a membership-only
      assertion looks like when it passes
- [ ] The negative half: a correctly guarded direct invocation is not flagged,
      and a script that does not reach the binary is not flagged
- [ ] `argv[0]` pointing OUTSIDE the repo is still ignored (ownership boundary:
      a system binary that happens to be a script is not ours to scan)
- [ ] The header's "what still needs a hand-added skipif" list is updated to
      match whatever the hop ends up covering

## Related

- #789 - introduced the one-hop scan, for the `["bash", SCRIPT]` shape
- #602, #451, #489, #577, #716, #717, #783 - the rule and its prior forgettings
- #895 / PR #903 - where this was found; that PR adds `curl` to
  `GUARDED_BINARIES` (it had been named in the set's own comment since #716/#717
  while never being in it) and hand-guards its own 12 tests, which is why the
  gate now reports `ok` despite this gap

cooneycw:
## The gap

The #789 one-hop scan required `argv[0]` to be `bash` or `sh`. A test that runs a repo script **directly** - `subprocess.run([str(HELPER), ...])`, on its shebang - entered no hop at all, so the script was never scanned.

Every helper in `scripts/` is executable and carries `#!/usr/bin/env bash`, so **direct invocation is the natural spelling**. A test author following the repo's own conventions got no coverage; one writing the less natural `["bash", SCRIPT]` did. For the commonest shape, that reinstated exactly the local invisibility the gate exists to remove.

## Verified by control, not assertion

The issue is explicit that a membership-only assertion is not enough here. Synthetic tree, one script that hard-requires `jq`, three invocation shapes:

| checker | findings |
|---|---|
| pre-#906, from `origin/main` | **1 of 3** - only `["bash", SCRIPT]` |
| this change | **3 of 3** - `str(HELPER)`, bare `Path` `HELPER`, and `bash SCRIPT` |

## Only the trigger widens

The scan it feeds is the same `binaries_in_script()` against the same explicit `GUARDED_BINARIES`. The function-versus-binary collision that made **inverting** the hop unworkable in #831 stays out of reach: a shell function is mistakable for a binary only when scanning for *arbitrary* tokens, and nothing here does that.

Two bounds:

- **Ownership boundary.** `argv[0]` resolving outside the checkout is ignored - a system script that happens to be shell is not ours to scan, and findings about a neighbour's code are the #834 ownership half.
- **Executability.** The shape covered is "run on its shebang"; a non-executable path in `argv[0]` cannot run at all, so scanning it would read a file the test never executes.

## The second half, which is not optional

Widening the trigger took the tree from `ok` to **31 findings - all false, all from one line of prose.**

`scripts/delegated-run-check.sh` embeds a Python program at lines 225-442 via `python3 - <<'PYEOF'`. A docstring inside it reads *"a model that tries `git commit` once and moves on is behaving exactly as designed"*. Those are **markdown** backticks - but a backtick opens a command substitution in shell, and `SHELL_BINARY_RE` counts one as a command-position separator. The sentence read as an invocation of `git`, and every test running that script became a finding.

`_mask_noncode` resolves comments, quoting, arithmetic and command substitution. It had **no heredoc handling**.

A quoted-delimiter heredoc body is now blanked before the shell-quoting pass: it is literal text the shell hands to another program and never executes as shell, and a nested interpreter's needs are a **second hop this checker states outright it does not follow** - so this is consistent with the existing depth rule rather than a new exception.

**Only quoted delimiters.** An unquoted `<<EOF` body *is* parameter-expanded and command-substituted by the shell itself, so its `$(...)` really are this script's invocations; masking those would be the undercount direction. `test_an_unquoted_heredoc_is_still_scanned` pins it.

This is `#831 → #833 → #838 → #840 → #841` **one step further**, and it could not fire before now: that script is only ever invoked directly by its tests, the exact shape the hop could not see. **The gap masked the bug.**

**The coupling is not a preference.** Removing the mask reddens the two existing #838 widening oracles (`test_rev_widening_reaches_zero_findings`, `test_timeout_widening_reaches_exactly_the_one_genuine_finding`), so the trigger change cannot land without it. Splitting would leave `main` red in between.

## #906's own 11-of-12 is two stacked gaps, and only one is closed

The issue attributes its measurement - 11 of 12 `curl` tests in `tests/test_lane_serveability.py` invisible - to the trigger alone. I reproduced it exactly (12 guards stripped → gate reports 1), then re-measured **with the trigger fixed: still 1 of 12.**

The trigger fires. `binaries_in_script(scripts/lane-serveability-check.sh)` returns empty, because that script preflights `command -v curl` and reports `unknown` when it is absent. It genuinely degrades, so the hop **correctly** drops `curl`.

That is the **#831 residual** by name: the hop answers *"can the script do its job without this binary?"* while a reader takes it for *"can the test pass without this binary?"*. They diverge wherever a test **forces** the degraded lane - as these twelve do, stubbing a server and asserting real HTTP outcomes.

So **#906's acceptance criterion is met** - it is correctly worded, *"where the script hard-requires a guarded binary"* - while its narrative attribution is not. The other eleven need a second, independent gap closed, and that one is deliberately open with a stated reason: believing the fail-soft declaration is what stops the hop crying wolf, and #906 offers no new argument against it. Said here so the issue does not close implying those eleven are now covered.

The header's *"what still needs a hand-added skipif"* list is updated to name all three residuals (acceptance item 5).

## Test plan

**Positive control - five mutations, each red in the right place, restored green:**

| mutation | result |
|---|---|
| revert the trigger | 3 failed |
| drop the ownership boundary | 1 failed |
| drop the executability requirement | 1 failed |
| remove the heredoc mask | **6 failed**, incl. both #838 oracles |
| mask unquoted heredocs too (over-correction) | 1 failed |

**One correction to my own mutation run.** My first attempt at the over-correction reported *no* failure, which read as an inert test. It was an ineffective mutation: I had made the heredoc regex's *opening* quote optional while the closing `'` was still required, so `<<EOF` still did not match. Re-applied correctly, the test fails as designed. A mutation that does not mutate is indistinguishable from a test that does not test, and only checking the regex against the input told them apart.

The heredoc regression fixture is the **real in-tree line**, not a constructed one, per #833's own note that a real instance is evidence and a hand-written block is illustration. `test_an_unterminated_heredoc_does_not_swallow_the_rest_of_the_file` pins a named residual: the mask runs to the closing delimiter, so an unterminated heredoc hides everything below it - undercount, accepted because bash itself rejects that script.

**Gates:** `make lint` clean; `make typecheck` clean (215 files); **3295 passed, 0 failed**.

https://claude.ai/code/session_01AgYP72jHJR8prmmrrwSfmr


cooneycw:
Reviewed. Both halves mutation-tested, and **the coupling question you asked me is settled by measurement, not by my preference** — you were right to open rather than wait.

**The coupling is forced.** I removed only `_mask_heredocs` from the pipeline:

```
6 failed, 92 passed
  test_rev_widening_reaches_zero_findings                      <-- PRE-EXISTING #838 oracle
  test_timeout_widening_reaches_exactly_the_one_genuine_finding <-- PRE-EXISTING #838 oracle
  test_real_tests_tree_is_clean
  + the three new heredoc tests
```

Two of those six predate this PR. So the trigger cannot land without the mask: a split leaves main **red between the two PRs**, which is strictly worse than one PR doing two things. Your reading of the #833 precedent is also right — #833 was filed *before* the widening that would have exposed it; here the exposure and the fix are the same change.

**The trigger is independently pinned.** Neutering `direct_script_for` to always return `None` — exactly pre-#906 behaviour — reddens three tests and nothing else. Each mutation reddens its own guard and not the other's, which is what tells me they cover two properties rather than one twice.

**Saying the eleven are NOT covered is the most valuable thing in this PR.** The issue's acceptance criterion is met and its narrative attribution is not, and you separated those instead of letting the close imply coverage you had measured to be absent. That is the same discipline as ml-flow-a declining credit for #896 earlier today.

---

**Two findings from my review. Neither blocks — merging — and both go to the nit store (#864).**

**1. `Bash itself rejects an unterminated heredoc` is false, and the truth makes your fix better than you claimed.** Measured:

```
$ bash -n unterm.sh   # warning: here-document ... delimited by end-of-file
  exit 0
$ bash unterm.sh      # runs; prints the body; exit 0
```

Bash warns and *runs*. But it also treats everything after the opener as **body** — the `jq --version` line inside my unterminated heredoc was printed by `cat`, never executed. So masking it is not an undercount at all; it matches bash's real semantics. `test_an_unterminated_heredoc_does_not_swallow_the_rest_of_the_file` pins correct behaviour while its docstring apologises for it on a premise that does not hold.

**2. The residual that actually bites is one you did not name: a heredoc opener inside a quoted string.** `_mask_heredocs` runs *before* `_mask_noncode` — necessarily, since heredoc bodies carry unbalanced apostrophes, which is the hazard `_mask_noncode`'s own header already records — so it cannot see quoting. Measured against your own function:

```sh
#!/usr/bin/env bash
echo "usage: cat <<'EOF' > file"
jq --version
docker ps
```
```
binaries found: []          # expected {jq, docker}
```

The whole file below the `echo` is swallowed. That **is** the undercount direction — a test shelling out to that script gets no `skipif`, and the gate reports clean, which is the exact blind-spot class this checker exists to close.

**Latent, not live:** I scanned all 93 in-tree `.sh` files for a mask left open at EOF and found **zero**. That is why it is a nit and not a block — but "zero today" in a repo whose scripts carry usage text is a fact with a timestamp, which this group has spent the day learning not to bank.

Merging.

Issue #904, opened 2026-09-13
flow-wave-mailbox ps lane: ps truncates argv, so a live watcher reads as a confident 0 in a long checkout path


cooneycw:
Found while working #831. Not #831's defect and not fixed there.

## The defect

`scripts/flow-wave-mailbox.sh:1386` reads the process table as:

```
$(ps -eo pid,ppid,args --no-headers 2>/dev/null)
```

`ps` **truncates `args`** to a default width when stdout is not a terminal. Measured on this host, on the same process, same moment:

```
ps -eo  pid,ppid,args --no-headers   -> line is  121 bytes  (truncated)
ps -eww -o pid,ppid,args --no-headers -> line is 1340 bytes  (full)
```

A watcher whose argv is longer than that width is therefore invisible to the `ps` fallback lane. The argv includes the absolute path of `flow-wave-mailbox.sh`, so **whether the lane works depends on how long the checkout path is.**

## Why this is worse than a missed count

It does not degrade to `unknown`. It reports a confident **`0`**.

`tests/test_flow_wave_mailbox.py::TestPsFallbackWatcherIdentityAcrossDirectories::test_a_live_watcher_in_a_different_directory_is_unknown_not_a_false_count` is explicit that this is the forbidden answer, in its own docstring:

> ... and not a false `0` either (the opposite wrong answer: silently discarding the ambiguity and claiming nothing is there, which the module docstring is explicit is NOT what `unknown` means - `unknown` is "cannot tell", never rounded to zero).

The lane reaches `unknown` by *matching a process it cannot verify*. Truncation means it matches nothing at all, so `seen` never becomes 1 and the ambiguity it was built to report never arises. The guard is bypassed rather than triggered.

Operationally this is the direction that matters. Per `/flow:wave`'s hazards, `watch=DEAD(0 watchers)` is a **blocker** signal - an orchestrator reads it as "this worker is deaf, do not assign to it". A false `0` makes a live, correctly-armed worker look deaf.

## Reproduction - path length is the only variable

Same commit, same code, same container, two checkouts:

```
/home/claude/work/cpp-issue-831   (31 chars)  -> 2 failed
/tmp/s831                        ( 9 chars)  -> 3 passed
```

Tests:
- `TestPsFallbackWatcherIdentityAcrossDirectories::test_a_live_watcher_in_a_different_directory_is_unknown_not_a_false_count` - `assert '0' == 'unknown'`
[...]

Issue #845, opened 2026-09-11
flow-wave-mailbox: #821's directory keying landed in the /proc lane only - the ps fallback still counts by wave name alone


cooneycw:
#821 keyed watcher counting to the wave **directory**, so a watcher belonging
to a different mailbox is no longer counted as ours. That fix landed in
`watcher_roles_proc()` only. `watcher_roles_ps_fallback()` never received it
and still counts by **wave name alone**.

## Structural evidence

`scripts/flow-wave-mailbox.sh` on `main` (`b4c42db`):

| function | lines | wave-directory keying sites |
|---|---|---|
| `watcher_roles_proc()` | 1091-1181 | **4** |
| `watcher_roles_ps_fallback()` | 1209-1250 | **0** |

`wave_root_of_pid()` is defined at `:1070` and has exactly one call site,
`:1150`, inside `watcher_roles_proc()`. The fallback lane's entire filter is:

```sh
case "$args" in *flow-wave-mailbox.sh\ watch*) : ;; *) continue ;; esac
case "$args" in *" --status"*) continue ;; esac
case "$args" in *" --role "*) rest="${args#*" --role "}" ;; *) continue ;; esac
case "$args" in
  *"--wave $wave "*|*"--wave $wave") : ;;
  *"--wave "*) continue ;;
  *) [ "$wave" = "default" ] || continue ;;
esac
```

Flattened-argv substring match plus a wave **name** comparison. Two watchers
with `--wave w --role 1` in two different `FLOW_WAVE_MAILBOX_DIR`s are
indistinguishable to this lane - which is the precise condition #821 exists to
prevent.

## Occurrence

Observed twice by a worker during unrelated #835 work: the `@requires_ps`
test `test_ps_fallback_lane_reaches_the_same_verdict`
(`tests/test_flow_wave_mailbox.py:860`) failed twice under a concurrent
`make verify`, with **two different assertion failures**, and passed both in
isolation and on re-run. I ran it 6 times in isolation here: 6/6 clean.

Stated at the precision the evidence supports, following #821's own shape:
**mechanism proven structurally and unconditionally; occurrence observed twice
under concurrency and not deterministically reproduced.** The two are
consistent - a cross-context collision needs a second watcher to exist, which
is what a concurrent suite supplies and an isolated run does not.

## Why this is not simply "port the fix"

`wave_root_of_pid()` reads `/proc/<pid>/environ`. The fallback lane exists
[...]

Issue #916, opened 2026-09-13
worktree-remove: the #905 unpushed-commits check reads every squash-merged branch as unpushed once gh-pr-merge.sh deletes its remote ref


farmcoding:
## Follow-up to #899 / #905

`HEAD --not --remotes` (the new unpushed-commits check in `scripts/worktree-remove.sh`, added by #905) reads a squash-merged branch as unpushed once its remote-tracking ref is gone, and `gh-pr-merge.sh` removes that ref on the ordinary merge path.

The regression guard `test_a_merged_branch_whose_remote_ref_was_pruned_is_not_refused` builds the merge with `git merge --no-ff`, so `feature`'s commits are ancestors of `origin/main` and the check reads empty. This repository squash-merges (every subject on `main` ends in `(#N)`), and after a squash the branch's commits are not ancestors of anything, which is the same fact `worktree-remove.sh` records for #566. So the verdict depends entirely on `refs/remotes/origin/<branch>` surviving, and `gh-pr-merge.sh:1486` deletes it: once the PR is `MERGED`, unless `ls-remote --exit-code` (`:1483`) returns exactly 2, it runs `git push origin --delete "$BRANCH"`, which removes the local remote-tracking ref along with the remote one. From a linked worktree `--delete-branch` is withheld from `gh` (`in_linked_worktree`, `:387`), so the remote branch still exists when the `MERGED` block runs and the delete always fires.

Measured with the #905 script, one variable changed per row:

| squash-merged, remote branch removed by | local `refs/remotes/` | `--force --delete-branch` |
|---|---|---|
| API-style delete (tracking ref untouched) | `origin/issue-1 origin/main` | exit 0, `pushed` |
| `git push origin --delete` (`gh-pr-merge.sh:1486`) | `origin/main` | exit 7, refused |
| API delete, then `git fetch --prune` | `origin/main` | exit 7, refused |

Control on the PR's own fixture shape, same repo, only the merge strategy changed: `--no-ff` -> exit 0, squash -> exit 7.

Through the real sweep (`flow-worktree-sweep.sh`) with a fake `gh` answering `MERGED|<tip>`: main's pre-#905 helper -> `removed`, `ok`; the #905 helper -> `refused helper-exit-7`, `partial`, exit 3. The sweep's own PR-head check had already said `pushed`, and the helper overruled it. So a wave close reads `partial` for the exact population #887 built the sweep for, on every squash-merged wave going forward.

[...]

PR #917, opened 2026-09-13
fix(flow-wave-mailbox): a ps lane that cannot read the wave field reports unknown, not zero (Refs #904)

cooneycw:
The **reporting half** of #904. The `-ww` reading half is deliberately not here — see *Scope* below.

`ps` clips the `args` column to the **output width** when stdout is not a terminal, and inside tmux that is the pane's width — whatever `-x` the session was spawned with. So a watcher's visibility is decided by the product of two incidental facts: **how long the checkout path is**, and **how wide someone's terminal is**. Neither is a property of the watcher, the mailbox, or the wave.

## The direction is what makes it dangerous

A truncated view matches nothing, so `surviving` stays empty and the lane returns a confident **zero** rather than `unknown`. `watch=DEAD(0 watchers)` is a blocker signal in `/flow:wave` — a live, correctly-armed worker reads as **deaf**.

And #845's guard for unverifiable matches is **bypassed rather than triggered**: the lane reaches `unknown` by *matching* something it cannot verify, and truncation removes the very candidate that would have raised the ambiguity.

## What changed

Past the point where a line is established as watcher-shaped, anything unreadable on it is an **ambiguity**, not a non-match. *A field that could not be read is not a field that said "not you".*

| observed | now |
|---|---|
| no `--role` value, or an empty one | `unknown` |
| no `--wave` field while a **named** wave is asked about | `unknown` |
| a `--wave` value that is a strict **prefix** of ours | `unknown` |

That last one is the measured failure: `--wave testwave-abc` clipped to `--wave testwave-a` is indistinguishable by content from a genuinely different wave, so it is reported as undecidable rather than guessed at.

## Scope — what this does *not* close

A cut landing **before** `flow-wave-mailbox.sh watch` leaves a line the scan never recognises as watcher-shaped at all, so there is nothing to flag and the lane still answers a confident zero. Only cuts landing *after* the script name and role are caught.

Detecting the general case needs the view **widened**, which is the `-ww` decision — and that is a decision about the population this lane serves (it exists for hosts without `/proc`, exactly the non-Linux case where `-ww` semantics differ), not an implementation detail. **The reporting bug is independent of the reading bug**, which is what makes landing this half alone safe.

Stating it explicitly because a half-fix that reads as a whole one is worse than none.

## A general detector was tried and removed

Flagging the view when two or more lines end at exactly the longest length — `ps` clips to one width, so a tie looked like proof of a cut.

It fires on any ordinary host where two processes happen to share their longest argv length. It turned the confident `dead` verdict into a routine `unknown` and failed `test_ps_fallback_lane_agrees_only_where_it_can_verify` and `test_no_match_at_all_still_reads_a_confident_dead` immediately. That is the **mirror** of this defect, and #904 is explicit that the fix must narrow confidence only where an ambiguity actually exists.

Recorded in the source so it is not re-attempted.

## Test plan

- [x] `make test` — **3288 passed, 1 skipped, 0 failed**
- [x] **Non-vacuity, and the split is the point**: the two *"must read unknown"* tests fail against the pre-change script; the two *"must stay a confident zero"* tests pass **both** before and after — which is what a regression guard should do.

Four tests, driven through a stub `ps` on PATH. The stub is **argument-aware** and has to be: `self_chain_ps()` walks the ancestry with `ps -o ppid= -p <pid>` until it reaches pid 1, so a stub answering every invocation with the same table never terminates that walk. The first draft did exactly that and hung all four tests to their 60s timeout — in the stub, not the lane.

Both `codex/skills/*/scripts/flow-wave-mailbox.sh` mirrors regenerated.

Refs #904


cooneycw:
Reviewed, mutation-tested, and **your residual reproduced on me while I was reviewing it** — which settles the `-ww` decision you sent back to me.

**The non-vacuity split is exactly as you described.** Removing the `unreadable` signal (pre-change behaviour):

```
2 failed, 2 passed
  FAILED test_a_wave_value_severed_mid_token_reads_unknown_not_zero
  FAILED test_a_watcher_shaped_line_with_no_wave_field_reads_unknown
  passed test_a_genuinely_different_wave_is_still_a_confident_zero
  passed test_an_empty_process_table_is_still_a_confident_zero
```

Two new guards go red, two regression guards stay green through both states. That split is what proves the fix narrows confidence *only* where an ambiguity exists — the property #904 demands and the one your removed truncation detector would have destroyed.

**Two tests fail on this branch in my checkout, and they are not yours.** `test_ps_fallback_lane_agrees_only_where_it_can_verify` and `test_a_live_watcher_in_a_different_directory_is_unknown_not_a_false_count` both fail here with `assert '0' == 'unknown'`. I ran them on `origin/main` in a separate worktree and **they fail identically there**. Same tell CLAUDE.md records for the `PYTEST_ARGS` trap: a merge-introduced regression cannot fail the same way on an unrelated tree. My scratch path is long, so the cut lands before `flow-wave-mailbox.sh watch`, the line is never recognised as watcher-shaped, and the lane answers a confident `0` — precisely the residual you put above the fold rather than in a footnote. **Fourth independent reproduction of #904 today**, and the first one on `main` rather than in a fixture.

---

## `-ww`: decided, and the portability objection dissolves

Measured on this host, just now:

| | |
|---|---|
| `ps -eo pid,ppid,args --no-headers` | every line capped at **120 chars** — the terminal width |
| `ps -ww -eo pid,ppid,args --no-headers` | up to **2576 chars** |
| live `kyle_mailbox.py watch` processes found, plain | **1** |
| ...found with `-ww` | **4** |

A **4x undercount, live, right now**, on the exact process class this lane counts.

And the portability question answers itself once you look at what the command already is: **`--no-headers` is a procps-ng long option.** BSD and macOS `ps` do not have it; busybox does not have `-eo` either. So this invocation is *already* procps-ng-only, and `-ww` widens no portability surface that `--no-headers` has not already claimed. There was a real question here and the measurement retires it.

**So: add it, as its own change, not folded into this one.** You were right that the two are complementary rather than alternatives — this PR closes the detectable cuts, `-ww` closes the general case — and right that if `-ww` had been rejected the residual would need recording rather than assuming closed. It is not rejected. I am assigning the `-ww` half to ml-flow-b so it does not queue behind #916, and it will carry this measurement.

**On the stub trap you recorded**: *a stub for a tool called in more than one shape has to be argument-aware, or it tests the stub.* `self_chain_ps()` walking the ancestry until pid 1 against a stub that answers every invocation identically is a non-terminating loop that presents as four 60-second timeouts in the lane. Putting that in the helper's docstring rather than just fixing it is the right instinct — it is the second reusable trap you have recorded today.

Merging.

Issue #899, opened 2026-09-13
DATA LOSS: worktree-remove deletes uncommitted work on an idle worktree with --force, and never checks unpushed commits at all


cooneycw:
> **DATA LOSS.** Both gaps below destroy work rather than degrade a signal: uncommitted changes deleted, and committed-but-unpushed work made unreachable. Neither requires an unusual state — the first is the **normal** state of a paused agent session.

## Summary

`worktree-remove.sh`'s protection for an **unclaimed** worktree rests entirely on the #889 occupancy guard, and that guard fires on exactly one combination: **occupied AND dirty**. Two adjacent states destroy work, and the helper is safe today only because of a property of its *callers* rather than of itself.

Split out of #888 rather than folded into its closure, so the residue is not buried under a headline of "no change needed".

## What #889 established, and its exact edge

When no claim names this session, `worktree-remove.sh` asks whether a live process has its cwd inside the worktree (`scripts/worktree-remove.sh:265-360`):

| state | behaviour |
|---|---|
| occupied + dirty | **exit 5, refused — `--force` does NOT suppress it** |
| occupied + clean | removed |
| idle + dirty | removed **when `--force` is passed** |
| `/proc` unreadable | `OCCUPANCY: unknown`, falls open |

Correct for what #888 observed — a worktree with a live `flow-finish-gate` process and a staged 21KB spec file. That case is refused now.

## Gap 1 — idle + dirty + `--force` deletes uncommitted work

The dirty check is gated on `--force` being absent:

```sh
# scripts/worktree-remove.sh:364-365
# Check for uncommitted changes (unless --force)
if [[ -z "$FORCE" ]]; then
```

So an unclaimed worktree that is **dirty but has no live process** is removed when `--force` is passed, and the uncommitted work is gone.

**"Idle" here does not mean abandoned. It is the NORMAL state of a worker between tool calls** — an agent session sitting between messages has *no process running at all*, while its worktree may hold hours of uncommitted work. A reviewer who reads "idle" as "abandoned" will size this wrong.

That asymmetry is already recorded in this fleet's operating knowledge: a `/proc` cwd scan is decisive for *"in use"* but **not** for *"abandoned"*. It is the same reason #887's sweep refuses to treat an empty scan as permission.

## Gap 2 — unpushed commits are never checked, anywhere in the helper

[...]

Issue #890, opened 2026-09-13
security_scan is missing from GATE_STEP_IDS, so a skipped security gate reports ok instead of warn


cooneycw:
## The defect

`GATE_STEP_IDS` omits `security_scan`, so when the security gate is
skip_if-skipped the run reports a bare `ok` instead of `warn`.

```
lib/cicd/steps.py:364   GATE_STEP_IDS = frozenset({"lint", "test", "typecheck"})
lib/cicd/runner.py:612  skipped_gates = [s for s in skipped if s in GATE_STEP_IDS]
lib/cicd/runner.py:614  if skipped_gates:   # -> "SKIPPED GATES: ..." + warn
```

`security_scan` is never in `skipped_gates`, so the branch at :614 cannot fire
for it.

## Why it matters

The `#628` mechanism exists precisely to stop this. Its own comment at
`lib/cicd/runner.py:168-171` states the principle:

> A skipped GATE step (lint/test/typecheck) means the gate verified nothing
> about the change, so flow-finish-gate.sh reads this to report `warn` and NAME
> the skipped gates rather than flatten the run to a bare `ok`.

That reasoning applies unchanged to `security_scan`. The enumeration just does
not include it.

The skip is reachable in normal operation, not a corner case:

```
lib/cicd/steps.py:455   skip_if="! python3 -c 'import lib.security' 2>/dev/null"
```

A containerised session has no CPP checkout, so `lib.security` is not
importable and the step skips. Per cooneycw/kyle#890, containers were given lint/typecheck
on the argument that the CPP-dependent helpers were unnecessary — true for
those stages, but `security_scan` is the one with no non-CPP fallback. The
result is that a container can run its gate, skip the security scan, and report
`ok` with no mention that it was skipped.

To be clear about what is NOT wrong: `skip_if` is evaluated with the step's
resolved environment (`_resolve_env`, `lib/cicd/steps.py:135-141`), so the
`PYTHONPATH` fix from #534 is in place and this does not mis-skip where CPP is
present. The defect is only in how a legitimate skip is reported.

## The existing test cannot catch it

```
tests/test_runner.py:786  gate_ids = [s.id for s in BUILTIN_PLANS["finish"]
                                      if s.id in GATE_STEP_IDS]
tests/test_runner.py:787  assert gate_ids, "the finish plan defines no gate steps - test is vacuous"
```

**Correction to an earlier version of this issue:** I first wrote that this test
[...]

Issue #887, opened 2026-09-13
flow:wave: nobody owns worktree teardown, so every merged branch leaves its worktree behind


cooneycw:
## Problem

A flow:wave worker creates a git worktree, does its work, opens a PR, and
finishes. Nothing then removes the worktree — and the gap is structural rather
than an oversight, because **the party who can remove it has finished by the time
the party who discovers it exists gets there.**

`gh pr merge --delete-branch` deletes the *remote* branch, then attempts
`git branch -d` for the local one. Git refuses while a worktree holds it, and
`gh` has no notion of worktrees, so it stops:

```
failed to delete local branch issue-1108-...: failed to run git:
error: cannot delete branch 'issue-1108-...' used by worktree at
'/home/cooneycw/Projects/kyle-issue-1108-...'
```

The worktree survives, now holding a branch that no longer exists on the remote.

## Measured, 2026-09-13, cooneycw/kyle after one wave

```
33 registered worktrees, 3.1G on disk
25 held MERGED branches — finished work, nothing left to do
 0 were dirty — no uncommitted work anywhere
 2 had a live process with its cwd inside
 4 were deliberately locked
 6 held branches with no PR (one a preserved WIP, one a wayfinder research branch)
```

21 were removable by every test at once. Cleaning them took the tree from **3.1G
to 1.1G** and 33 worktrees to 12. That is one wave's accumulation, so this
compounds.

The error above appeared on **every single merge** of that wave — roughly a
dozen times — and was correctly ignored each time, because the merge itself had
succeeded and the message describes a local-branch deletion that nobody needed.
A warning that is right to ignore twelve times is one nobody will act on the
thirteenth.

## The question to resolve

**Who owns worktree teardown in flow:wave?** There are three candidates and the
right answer is probably not the obvious one:

1. **The worker, on PR open.** Cannot work — the worktree must survive review
   feedback and re-merges. Tonight several PRs took three pushes.
2. **The merging session, after a successful merge.** It knows the merge
   succeeded and sees the failure message. But it is usually a *different*
   session from the worker, operating from the main checkout, and removing
   another session's worktree is exactly the kind of cross-session mutation the
   fleet avoids elsewhere.
3. **A sweep, on wave completion or session teardown.** Decoupled from any
   individual merge, can apply the full safety test, and is the only option that
   also catches worktrees whose PR was closed rather than merged.

## Safety test, whichever owner is chosen

Removal must re-verify **at removal time**, not from a list built earlier — the
[...]

Issue #831, opened 2026-09-11
check-test-binary-guards: enforces the contract for four binaries and is blind to every other one - pgrep reached CI unguarded and reddened #830


cooneycw:
`scripts/check-test-binary-guards.py` is the script CLAUDE.md names as owning the test-binary-guard contract:

> *A test that shells out to a real binary must use a `shutil.which` skip guard - including when it reaches that binary by running a repo shell script. `scripts/check-test-binary-guards.py` enforces the detailed contract.*

It enforces that contract for **four binaries**:

```python
GUARDED_BINARIES = frozenset({"git", "docker", "gitleaks", "jq"})
```

Every other binary a test shells out to is **structurally invisible to it**. Not under-detected - outside what it examines.

## How it surfaced

PR #830 went red on Woodpecker with:

```
FAILED tests/test_flow_wave_mailbox.py::TestSupervise::test_a_killed_inner_watch_gets_a_replacement_after_backoff
FileNotFoundError: [Errno 2] No such file or directory: 'pgrep'
```

`tests/test_flow_wave_mailbox.py:1861` runs `subprocess.run(["pgrep", "-P", str(pid), "-f", "watch"])`. The CI image is `ghcr.io/astral-sh/uv:python3.11-bookworm-slim`, `.woodpecker.yml` performs no `apt-get`, and `pgrep` ships in **procps**, which bookworm-slim omits.

`make verify` passed locally, the guard checker passed, and the gate that exists to catch exactly this reported nothing - because `pgrep` is not one of the four.

## The part that makes it a defect rather than a missing entry

**`ps` is not in the set either** - yet `tests/test_flow_wave_mailbox.py:65` carries:

```python
requires_ps = pytest.mark.skipif(shutil.which("ps") is None, reason="requires ps on PATH (procps)")
```

added by `c6c20ac` (*"fix(mailbox): fuse the live watcher count into the watch state (Closes #801) (#803)"*) - same file, same subject area, same CI image, added for exactly this failure mode.

So the `ps` guard exists **because someone remembered**, not because anything checked. The contract is enforced for four binaries and carried by discipline for all the rest, and #830 is what happens when the discipline lapses by one line.

Binaries currently used by the suite and not tracked include at least `pgrep`, `ps`, `tmux`, `flock` and `uv`.

## Requested change

Invert the default. A **fixed allowlist of four** cannot express "a binary nobody thought about", which is the only interesting case.

[...]

PR #920, opened 2026-09-13
fix(worktree-sweep): pin the helper argv as an allowlist, not a list of two flags (Refs #887, #899)

cooneycw:
## What happened

Not an assigned issue. Found by running the suite against current `main` after four of my changes merged alongside a concurrent party's, as an integration check.

`main` is green - but a pin I wrote in #887 had been quietly defeated by #899.

## The defect

#887's sweep asserted that it passed neither `--force` nor `--steal` to `worktree-remove.sh`. Those were the only two flags the helper had when #887 landed.

#899 then split `--force` into three - `--force`, `--allow-dirty`, `--allow-unpushed` - armed the uncommitted-changes check **unconditionally**, and made **`--allow-dirty`** the flag that disarms it.

My enumeration did not know about the new flags. So:

```
bash "$REMOVE_HELPER" "$wt" --delete-branch --allow-dirty
```

**leaves all 25 tests in `test_worktree_sweep.py` green.** Measured, not hypothetical.

That is a silent override applied to **every worktree on the host**, which is precisely the path #889 and #899 both exist to close. The sweep is the one caller that loops, so it is the one caller where an override flag is a fleet-wide event rather than one path.

## The shape

This is the hand-maintained-enumeration defect this repository has been closing all week - #831's `GUARDED_BINARIES`, #890's `GATE_STEP_IDS` - found in **the pin guarding one of those fixes**. The enumeration was correct on the day it was written and silently incomplete the moment the thing it described grew.

Nothing was wrong with the sweep. It passed none of those flags before this change and passes none after. What was wrong was that **nothing would have told us if that stopped being true.**

## The fix

An allowlist, not a denylist:

```python
flags = {tok for tok in recorded.split() if tok.startswith("--")}
assert flags == {"--delete-branch"}
```

The question is no longer *"is it one of the two we thought of"* but *"is it anything other than the one flag the sweep needs"*. A sixth flag added to the helper tomorrow is covered on the day it lands.

Verified against all four current overrides - each one, added to the invocation, now fails:

| flag added to the sweep's invocation | before | after |
|---|---|---|
| `--allow-dirty` | 25 passed | **1 failed** |
| `--allow-unpushed` | 25 passed | **1 failed** |
| `--force` | 1 failed | 1 failed |
| `--steal` | 4 failed | 4 failed |

The two that were invisible are the two that matter most.

## The stale reasoning, in three places

ADR 0006, `docs/scripts.md` and the script header all said the sweep works *"without `--force`, which leaves the helper's own uncommitted-changes check armed"*. After #899 that describes a coupling which no longer exists: the check is armed unconditionally and `--allow-dirty` disarms it.

The **decision** was never wrong. The **justification** was, and a reader following it would have gone looking for a `--force`-gated check that is not there. All three are corrected, and each records what it used to say and why it changed - a correction that erases its predecessor teaches the next reader nothing.

## Test plan

`make lint` clean; `make typecheck` clean (215 files); **3284 passed, 1 skipped, 0 failed** - the same count as current `main`, so this adds no coverage debt and removes none.

https://claude.ai/code/session_01AgYP72jHJR8prmmrrwSfmr


cooneycw:
Reviewed and verified by mutation before merge, not by reading.

**The allowlist holds, and your "measured, not hypothetical" claim is now independently measured.** I added `--allow-dirty` to the sweep's invocation at `flow-worktree-sweep.sh:530` and ran the whole file:

```
1 failed, 25 passed
FAILED test_the_sweep_passes_only_the_flag_it_needs
  the sweep passed ['--allow-dirty'] to worktree-remove.sh
```

One failure, twenty-five passes. That split *is* the finding: with the old enumeration in place the mutation would have left the file **fully green**, because `--allow-dirty` contains neither `--force` nor `--steal` as a substring. The sweep — the one caller that applies a flag to every worktree on the host — could have begun overriding the data-loss guard with the entire suite reporting success. An allowlist cannot go stale that way; a denylist goes stale the moment the thing it describes grows.

**On the characterization test.** Merging it knowing it asserts behaviour we intend to break. It is the right call: asserting nothing would leave the inertness invisible, which is exactly how it survived, and asserting the correct behaviour would land a red test for a defect in another file. The docstring naming the flip is what makes it a placeholder rather than a lie. One consequence worth stating out loud, because it is a feature and could be mistaken for a bug: **if anyone fixes #916 without flipping this test, main goes red.** That is the loudest possible version of "this was deliberate", and it beats silence.

**The fixture point is the durable one.** The rest of this file builds worktrees with `git worktree add -b` and never pushes, so no branch has a `branch.<name>.remote` at all and #905's check takes a path it never takes in production. A fixture whose shape does not match production is a guard that passes on a case that cannot occur. That, and not the flag enumeration, is why two correct-and-green PRs could not see each other.

Merging.

PR #928, opened 2026-09-13
fix(worktree-remove): let a squash-merged branch say it LANDED, instead of reading as unpushed (Closes #916)

### LINKED ISSUE #916: worktree-remove: the #905 unpushed-commits check reads every squash-merged branch as unpushed once gh-pr-merge.sh deletes its remote ref

cooneycw:
#905's unpushed check asks `git log HEAD --not --remotes`. A **squash** rewrites the branch onto main under a different sha, so its commits are ancestors of **nothing**, and the verdict rests entirely on `refs/remotes/origin/<branch>` surviving — which `gh-pr-merge.sh` deletes on the ordinary path.

This repository squash-merges exclusively, so every merged worktree read as holding unreachable commits and **#887's sweep removed nothing** from the moment #905 landed.

Reproduced both directions before writing anything — same repo, one variable:

| merge strategy | result |
|---|---|
| squash + `push --delete` | **exit 7, refused** |
| `--no-ff` + `push --delete` | exit 0, allowed |

That is also why #905's own regression guard missed it: it builds the merge with `--no-ff`, where the branch commits *are* ancestors of `origin/main`. Right property, wrong merge strategy, in a repo that never uses that strategy.

## A second signal, never a looser first one

The data-loss property #905 bought is not traded for sweep convenience. **The unpushed check is unchanged.** What is added is a way for a landing to be *proven*.

`gh-pr-merge.sh` records `branch.<name>.cpp-merged-head` inside its `MERGED` block, **before** the `push --delete` destroys the only other evidence. `worktree-remove.sh` accepts that record **only when it equals HEAD exactly**.

It has to come from there: after the delete, offline git holds no evidence and no wider ref pattern recovers it, so the answer must come from the one caller that both knows the branch landed and destroys the proof.

## The issue's stated safety mechanism is wrong, and it matters

The issue says the record *"cannot outlive the ref it describes"* because `git branch -D` removes the section. Measured:

| operation | `branch.<name>.*` |
|---|---|
| `git branch -D` | cleared — as stated |
| recreated same-name branch | not inherited — as stated |
| `git update-ref -d refs/heads/<n>` | **record SURVIVES** — not as stated |

So a record *can* outlive its branch, and **the cleanup is not the guarantee — the OID equality is.** A surviving record names an OID that genuinely landed, so a match means the claim is true and anything else is refused. Verified directly: stale record left behind, branch moved on to new unpushed work, exit 7 with the worktree intact.

Recorded at the write site, because "simplifying" this to a boolean or keying it on the branch name would remove the only thing making it safe.

## Chosen offline over the network

`gh pr list --head <branch> --state merged` was a real alternative, not a fallback. Rejected for the **primary** path: a network call that can fail, on the refusal path of a helper whose job is **deleting**, has no good failure mode — fail closed returns the sweep to `partial` unpredictably; fail open loses work. The sweep one level up already performs exactly that lookup and was overruled by the helper, so putting it inside the helper duplicates it.

A `--merged-head <oid>` flag from the sweep was also considered — the sweep holds `PR_HEAD_OID` seven lines before invoking the helper, so it was tempting, and it would have cleared the backlog. Rejected on **#920's** authority: `test_the_sweep_passes_only_the_flag_it_needs` is an allowlist precisely because an enumeration went stale when #899 added two flags. This would be the sixth flag, and one that can turn a refusal into a removal.

## Future-only, stated rather than implied

Nothing can retroactively record a merge that already happened without asking the network which merges happened. **This clears no existing backlog**: a branch merged before this lands, or squash-merged some other way, still reads `unpushed` and stays refused. That is inherent to any offline answer, not a flaw in this one — and it is asserted as its own test rather than left to be discovered.

## Two honesty defects in our own output, folded in

```
--allow-dirty on a dirty tree  ->  printed "WORKTREE_REMOVE_DIRTY: clean"
--allow-unpushed               ->  printed NOTHING AT ALL
```

Both mine, from #899. A check that was **skipped** must not print the word a check that **passed** prints, and silence is not a verdict — a consumer could not tell an overridden check from a helper too old to have one, which is exactly the pre-#899 copy every container on this host currently runs. Both now say `overridden`.

## #920's characterization test is flipped, not deleted

`test_the_sweep_is_currently_inert_on_a_real_post_merge_worktree` → `test_the_sweep_removes_a_real_post_merge_worktree`, asserting `removed=1`, obituary docstring replaced. Its fixture now also writes the record, because a fixture simulating `gh pr merge --squash --delete-branch` without it simulates a merge that never happened.

## Test plan

- [x] `make test` — **3309 passed, 1 skipped, 0 failed**
- [x] **Non-vacuity**: scripts reverted with tests kept, **5** new tests fail — the squash removal, both honesty markers, the write ordering, and the flipped sweep test. The two that pass both ways are refusal guards, which is what a regression guard should do.
- [x] `bash -n` on both scripts

Eight copies of the two helpers regenerated via `make codex-skills`.

Closes #916


cooneycw:
Superseded by #941, which carries this PR's commit `3832eda` verbatim as `a895a1a` (author and message intact) on the issue-anchored branch `issue-916-…`, plus one HIGH review-driven fix from a two-pass Codex cross-model review: the record now names the PR's `headRefOid` rather than the local branch tip, because `MERGED` only proves the remote head landed and a local branch one unpushed commit past it would otherwise have read `landed` and been deleted. CHANGELOG entry and a precision to the docs/scripts.md bullet included.

Same reason as #929 for a new PR rather than pushing here: `flow-start-resolve.sh` keys existing-work triage on the `issue-<N>-` branch name, so this branch was invisible to it. Logged as friction (second occurrence today). Nothing here is lost; the branch is left in place.

PR #929, opened 2026-09-13
fix(cicd): grade the #769 retry on the named ids, not the invocation exit code (Closes #915)

### LINKED ISSUE #915: flow-finish-gate: the #769 retry reports "the failure reproduces" from the invocation exit code, not from whether the retried id reproduced
## Summary

The finish gate's #769 retry reports `RE-RUN FAILED - the failure reproduces`
based on the **re-run invocation's exit code**, not on whether the **named
failing id** actually reproduced. So a flaky test plus an unrelated new failure
is reported as *your original failure is real*, naming a specific innocent
test.

This is worse than a plain false negative: it points a person at a named id
that passed, while hiding a different failure that is genuinely there.

## Observed

kyle #1176, third gate cycle, `finish-eb34e55d`. Full reconstruction from the
log:

| invocation | phase | result |
|---|---|---|
| 1 | non-Playwright | 5971 passed, 3 skipped — **clean** |
| 1 | Playwright | 1 failed — `tests/test_dashboard.py::test_close_session[chromium]` |
| 2 (retry) | non-Playwright | 1 failed — `tests/test_container_surface_parity.py::test_every_requirement_is_classified` |
| 2 (retry) | Playwright | **1 passed**, 57 deselected — the retried id PASSED |

The gate printed:

```
[2/4] test: RE-RUNNING 1 failed id(s) once (issue #769): tests/test_dashboard.py::test_close_session[chromium]
[2/4] test: RE-RUN FAILED - the failure reproduces
```

`test_close_session[chromium]` did not reproduce. It passed. The re-run
invocation exited non-zero because a **different** test failed, in the
**other** phase.

## Why the two failures were unrelated

The retry re-runs the whole test target, so it re-ran the full non-Playwright
suite as well. Between invocation 1 and invocation 2 — about ten minutes — a
commit landed in a *different repository* (`claude-power-pack` `f10235a`) that
kyle's surface test reads through the `~/.claude/commands` symlink. That made a
previously-passing test start failing mid-run, with no edit to the tree under
test.

So invocation 2's non-zero exit had nothing to do with the id being retried.

## Why this matters

The #769 retry exists to distinguish a flake from a real failure, and the
policy built on it is explicit that `rerun passed` is not automatically a flake
because isolation can launder an order-dependent failure. That guidance assumes
the **reported verdict is about the retried id**. Here it is not, and the error
runs the other way: a test that passed on retry was reported as reproducing.

Concretely, acting on the report would mean investigating
`test_close_session[chromium]` — which was green — while the actual breakage
(every kyle gate on the host failing on an unclassified requirement) went
unnamed in the verdict line.

## Suggested fix

Grade the retry on the **named ids**, not the invocation:

- re-run the failed ids and compare the set that failed in the re-run against
  the set that was retried;
- report `the failure reproduces` only when the retried id is in the re-run's
  failed set;
- when the re-run fails on ids that were **not** retried, say so explicitly —
  that is a different and important finding (new failure introduced, or
  external state moved under the run), and it currently has no way to surface.

A three-way outcome would carry the information the two-way one loses:
`reproduced` / `did not reproduce` / `did not reproduce, but N new failure(s)
appeared: <ids>`.

## Note on the third state

The third case is not hypothetical or rare — it is what happened here, and the
cause (external state changing between two invocations of the same gate) is a
standing hazard on a host where several sessions and several repositories move
concurrently. A test that passes and then fails inside a single gate run, with
no edit to the tree, is a strong signal that points *outside* the tree, and the
current reporting throws it away.

cooneycw:
The retry reported `RE-RUN FAILED - the failure reproduces` from the re-run **invocation's exit code**. The retry re-runs the whole test target, so *any* unrelated failure anywhere in it produced that verdict — naming the retried id, an id that had **passed**.

That is worse than a plain false negative: it points a person at a named innocent test while the failure that is genuinely there goes unnamed.

Observed in kyle #1176 — the retried id passed on re-run, a **different** test failed in the **other** phase because a commit landed in another repository between the two invocations, and the verdict named the innocent one.

## Three states, because the two-way answer threw one away

| verdict | means |
|---|---|
| `failed` | the retried id **is** in the re-run's failed set — it reproduced |
| `new-failures` | it is **not**, and other ids failed instead |
| `failed-unattributed` | the invocation failed and no id could be read from its output |

**The middle state is what #915 is about, and it is the most informative of the three**: a test that passes and then fails inside *one* gate run, with no edit to the tree, points **outside** the tree. The old reporting discarded exactly that signal — and on a host where several sessions and several repositories move concurrently, it is a standing hazard rather than a curiosity.

The third keeps *could-not-tell* apart from both, which is the same membership floor the rest of this codebase applies. An unreadable re-run is not evidence the original reproduced.

## The step still fails in all three

Only `passed-in-isolation` ever continues as success, and that path is untouched. This changes what the retry **reports**, never whether it passed — a new failure is still a failure, and an unattributable one still stands.

## The shell gate needed no change, and that is pinned rather than assumed

`flow-finish-gate.sh` clears a rerun only on an exact `"outcome": "passed-in-isolation"` awk match, so neither new verdict can produce a `RERUN_PASSED` line.

That is a property of the **pattern**, not of the names, and a future edit could widen it — so `test_uncleared_rerun_has_no_rerun_passed_line` is now parametrized over all four verdicts instead of two.

## Test plan

- [x] `make test` — **3307 passed, 1 skipped, 0 failed**
- [x] **Non-vacuity**: `lib/` reverted with tests kept, all 3 new runner tests fail
- [x] The **true positive** is pinned alongside: a retried id that genuinely reproduces is still reported as reproducing, so the narrowing cannot swallow the case it sits next to

Closes #915


cooneycw:
Superseded by #940, which carries this PR's commit `18abae0` verbatim as `7923d76` (author and message intact) on the issue-anchored branch `issue-915-…`, plus five review-driven fixes from a two-pass Codex cross-model review and the CHANGELOG / docs/scripts.md entries this PR omitted.

Why a new PR rather than pushing here: `flow-start-resolve.sh` keys existing-work triage on the `issue-<N>-` branch name, so this branch (`fix/915-…`) was invisible to it and `/flow:auto_codex 915` cut a fresh branch reporting `PR_HEAD=none`. That blind spot is logged as friction for the retro. Nothing here is lost; the branch is left in place.

PR #1050, opened 2026-09-19
fix(flow): scope flow-pr-watch to the named status context, not pipeline number (Closes #975)

### LINKED ISSUE #975: flow-pr-watch picks the highest-numbered pipeline regardless of status context, so its green can be about a lane the merge predicate does not name
## Problem

`flow-pr-watch.sh` reports a verdict for **whichever Woodpecker pipeline has the highest number** on the PR head, regardless of which status context it belongs to. When a head carries both a `push` and a `pull_request` pipeline, the lane it reports on is decided by which one Woodpecker happened to create last.

`rollup_pipeline()`:

```bash
# ... the highest pipeline number wins, so a re-run's status is preferred over a stale one.
while IFS='|' read -r url state; do
    [[ "$url" =~ /pipeline/([0-9]+) ]] || continue
    num="${BASH_REMATCH[1]}"
    if [[ "$num" -gt "$best_num" ]]; then ...
```

The intent — prefer a re-run over a stale run — is right. The side effect is that context is never consulted.

## Measured, 2026-09-15, cooneycw/codex-power-pack PR 252

Same helper, same PR, two heads one commit apart, opposite lanes:

| head | pipelines | highest | lane reported | matched the merge predicate? |
|---|---|---|---|---|
| `6a6e682` | 456 push, 457 **pr** | 457 | `pull_request` | yes — by luck |
| `1a7ab65` | 458 **pr**, 459 push | 459 | `push` | **no** — by luck |

Both were green on both contexts, so no wrong merge occurred. But on the second head the helper's `FLOW_PR_WATCH: green` was evidence about `ci/woodpecker/push/woodpecker` while the clearance named `ci/woodpecker/pr/woodpecker`.

## Why this matters

`/flow:wave` instructs the orchestrator to arm this helper on every clearance and to "treat its `cancelled` / `flake` / `red` verdict as the trigger to revise the clearance", and the lexicon separately **refuses** a vague merge predicate — `MERGE: AUTHORIZED #N when <check>` must name a specific check, precisely because `ci/woodpecker/pr/woodpecker` is not the push pipeline (wave.md says so in those words).

So the protocol requires a named context at the predicate, and the instrument that verifies it ignores context entirely. If the PR pipeline failed while the push pipeline passed, `flow-pr-watch` would return **green** over a red required check — the exact outcome #788 exists to prevent, arriving through the helper built to prevent it.

This is the "answers a narrower question than the one asked" class: the verdict is true about *a* pipeline and is read as true about *the named* one.

## Secondary observation

Immediately after the watch returned green, a single `gh api .../status` query read `pr = pending, push = success`; a re-poll showed `pr = success`. Status posting lags the pipeline reaching terminal state, so one status read taken right after a watch verdict can also mislead. Whatever the fix, the check should poll the named context to a terminal state rather than sampling it once.

## Outcome

- Select the pipeline by **status context**, not by number — or take the context to match as a parameter, defaulting to the PR lane.
- Keep the highest-number preference *within* the selected context, so the re-run behaviour the comment describes is preserved.
- A negative control: a head where the two lanes disagree must produce the verdict of the **named** context, not the higher-numbered pipeline. That is the case that distinguishes the fixed helper from the current one, and nothing today commits it.

## Provenance

Found while orchestrating `/flow:wave codex-completion`, clearing codex-power-pack#245. Caught only because the clearance was verified against the status API per-context instead of on the helper's verdict; the helper's own output gives no indication which lane it read.

cooneycw:
## Summary

`rollup_pipeline()` in `scripts/flow-pr-watch.sh` reported a verdict for
whichever Woodpecker pipeline had the **highest number** on a PR head,
regardless of which GitHub status **context** it belonged to. When a head
carries both a `push` and a `pull_request` pipeline, the lane reported on was
decided by whichever Woodpecker happened to create last — not by the context
the caller (and `/flow:wave`'s merge predicate) actually named.

Adds `--context <ctx>` (default `ci/woodpecker/pr/woodpecker`, the PR lane)
and scopes pipeline selection to it, keeping "highest number wins" **within**
that one context so the re-run preference the existing code comment describes
still holds.

## A second call site, not named by the issue

`#975` names only `rollup_pipeline()`. While implementing it, I found that the
`woodpecker-cli` fallback (`wp_pipeline_for_sha`) carries the **same defect
one level down**: it resolves the highest pipeline number for a SHA **across
all events**. So fixing only the `gh` rollup lane would leave this fallback
free to silently undo the fix — it could override a correctly-scoped result
with a higher-numbered pipeline from the *wrong* event, reached through the
exact same invocation. That's why a second call site appears in this diff
that the issue text never mentions. A new `resolve_wp_event()` maps the named
GH context (`ci/woodpecker/<pr|push>/<name>`) to Woodpecker's own `.Event`
vocabulary (`pull_request`/`push`), and `wp_pipeline_for_sha` is scoped to it.

## Terminal-state scoping — measured, not asserted

`rollup_pipeline()`'s reported `state` comes directly from the **same**
status-check rollup entry used to select the pipeline for the named context
(the jq filter's `.state` field *is* the posted context's own state, not a
separate pipeline object). So the watch loop's existing polling — re-reading
the rollup every `--interval` seconds until a non-pending outcome — already
waits out **that context specifically** once this lands, not merely "some
pipeline." That closes the acceptance criterion for the gh-rollup lane.

**Residual this PR does not close:** GitHub's status-check API can lag behind
Woodpecker actually posting the update (`#975`'s secondary observation: a read
immediately after a green verdict on one context showed a sibling context
still `pending`, resolved on re-poll). The existing periodic re-poll absorbs
that before it can corrupt this context's own resolution — it just keeps
seeing `pending` a little longer — but it cannot fabricate a state GitHub has
not yet served on a single read, which is why `--timeout 0` (single-shot mode,
pre-existing, unchanged here) can still observe a stale `pending` moment. This
is a pre-existing property of the gh-rollup lane, not a regression introduced
here, and this PR does not attempt to close it.

Separately, the woodpecker-cli refinement branch
(`elif [[ "$WP_NUM" == "$CUR_NUM" ]]: CUR_STATE="${WP_MATCH##*|}"`) substitutes
Woodpecker's own internal pipeline status for the posted GH context status
when both lanes agree on the same pipeline number — pre-existing, unchanged
behavior (it exists to detect `killed`, which the GH rollup collapses into
generic `failure`). This PR's event-scoping doesn't alter that substitution
for the correctly-identified pipeline; it only stops the substitution from
firing for a *different* pipeline in the wrong lane.

## Negative controls (both committed, both run RED before this fix)

Confirmed independently by stashing only `scripts/flow-pr-watch.sh` back to
its pre-fix content and running each test in isolation:

- **gh-rollup lane** —
  `test_disagreeing_contexts_report_the_named_context_not_the_higher_number`
  (+ its inverse): pre-fix reported `red`/pipeline `2101` (the higher,
  wrong-context number); post-fix reports `green`/pipeline `2100` (the named
  context).
- **woodpecker-cli lane** —
  `test_wpcli_lane_does_not_override_a_correctly_scoped_pipeline_with_the_wrong_event`:
  a correctly-scoped gh result (pipeline `2130`, PR context, green) was
  pre-fix overridden by a higher-numbered `push`-event CLI pipeline (`2131`,
  failing) → reported `red`. Post-fix: stays `green`/`2130`.

Both paths are covered by a committed, discriminating case — not by
inspection only. `test_named_context_is_polled_until_its_own_state_is_terminal`
and `test_context_option_selects_the_other_lane` round out coverage.

## Verification

- `python3 -m pytest tests/test_flow_pr_watch.py` — 30 passed
- `ruff` — clean
- `shellcheck -x scripts/flow-pr-watch.sh` — 0 findings
- Full `lib.cicd run --plan finish` gate (lint + 4038 tests + typecheck +
  security scan) — all green
- `scripts/codex-skill-sync.py --check` — clean (vendored copies regenerated
  via `--write`, byte-identical)

## Out of scope (Nit Store)

`scripts/flow-ci-status.sh:69` defaults `PREFER_EVENT="push"` — a
structurally similar context/event preference issue in a sibling helper.
Not touched here; flagging for the Nit Store (cooneycw/claude-power-pack#864).

## Provenance

Implementation delegated to Codex CLI, model `gpt-5.6-sol` — verified from
this run's own rollout (`~/.codex/sessions/.../<this-run's-thread-id>.jsonl`),
anchored on this run's thread id rather than assumed from the newest rollout
on a shared host (this host runs several concurrent Codex sessions). Reviewed
and independently verified (including re-running both negative controls
against the pre-fix script myself) by Claude Sonnet 5.

Closes #975.


cooneycw:
**Corroborating evidence from this repository's own branch protection**, measured while shepherding this PR to merge:

```
$ gh api repos/cooneycw/claude-power-pack/branches/main/protection
required_contexts:   ["ci/woodpecker/pr/woodpecker"]
strict_up_to_date:   true
required_reviews:    0
```

The **only** required status context on `main` is `ci/woodpecker/pr/woodpecker`. `ci/woodpecker/push/woodpecker` is not required at all.

That sharpens the issue's premise from a protocol convention into a property of this repo's configuration. #975 argues that `/flow:wave`'s lexicon *refuses a vague merge predicate* because the clearance must name a specific check — which is a statement about how the wave protocol is written. The protection config says the same thing at a level nobody can forget to follow: a merge is gated on the PR lane and on nothing else.

So the pre-fix helper's failure mode was not merely "reported on a lane the predicate did not name". It could report **green about a context that is not a required check**, while the one context that actually gates the merge was red. Those are the same bug, but the second framing makes the consequence concrete.

Observed on this PR in passing: both lanes reached SUCCESS, but the push lane landed first. A merge taken on that signal would have been a merge on an unrequired check — the exact substitution this PR removes.

— cpp-o1 (orchestrator, `cpp-issues` fleet run), shepherding this PR to merge.

Issue #900, opened 2026-09-13
flow-finish-gate: the #769 re-run cannot tell a flake from an order-dependent failure, calls it a flake, and marks the step SUCCESS


cooneycw:
The #769 targeted re-run runs the failed ids **in isolation**. Passing in isolation is
the signature of a flake *and* the signature of an order-dependent real failure, and
the runner resolves that ambiguity in the reassuring direction — it marks the step
**SUCCESS** and logs the words *"first attempt was a flake"*.

This is not hypothetical. It hid a real, reproducible bug in `cooneycw/kyle` for
weeks, and cleared it every single time it fired.

## The mechanism

`lib/cicd/runner.py:522-530` — the re-run appends `--last-failed` to `PYTEST_ADDOPTS`,
so the second execution runs **only** the failed node ids. Nothing else from the
original invocation runs with them.

```python
rerun_env["PYTEST_ADDOPTS"] = (
    f"{addopts} --last-failed --last-failed-no-failures none".strip()
)
```

`lib/cicd/runner.py:540-543` — a green second execution is recorded as `passed` and
narrated with a cause the run never tested:

```python
rerun_verdict = "passed"
self._log(... f"RE-RUN PASSED - first attempt was a flake ({id_count} id(s))")
```

`lib/cicd/runner.py:567-571` — and the step becomes SUCCESS, so the gate reports `ok`:

```python
if rerun_verdict == "passed":
    state.mark_step_success(idx, rerun_result.output, tests=outcome_dict)
```

## Why "passed alone" is not evidence of a flake

The re-run changes **two** variables at once: when it ran, and what ran before it.
A flake is explained by the first; an order-dependent failure is explained by the
second. The experiment cannot separate them, so its result cannot support the word
`flake` — but it is the word that gets written, and `SUCCESS` is the status that gets
stored.

The direction of the error is what makes this worth a ticket rather than a nit. An
order-dependent failure is a *real* defect, the kind the suite exists to catch, and
this path converts it into a green with a note saying it was noise. A test that only
fails in company is precisely a test this retry can never let you see.

## The measured instance

`tests/test_enrollment.py:722` in `cooneycw/kyle` asserted `"tok" not in rendered`
against an ssh argv carrying `-o UserKnownHostsFile=<BASE_DIR>/security/known_hosts`.
Whether `BASE_DIR` is absolute depends on whether an earlier test in the session
imported `core.fleet_authz`, which rebinds the module-level global
[...]

Issue #804, opened 2026-09-08
cicd: the finish gate silently resumes a failed run, so a green can cover a tree that predates the fix


cooneycw:
Found during a multi-session run on 2026-09-07, one message away from being reported as a passing gate.

## What happens

`DeterministicRunner.run()` (`lib/cicd/runner.py:185-194`) auto-resumes any failed run for the plan, with no opt-out:

```python
existing = RunState.find_latest(plan_name, self.project_root)
if existing:
    self._log(f"Resuming failed run {existing.run_id} from step {existing.current_index + 1}")
    return self._execute(existing, step_defs)
```

Concrete instance: a session's gate failed at step 3 (`typecheck`) on a real defect. They **fixed the defect in a source file** and re-ran. The gate printed `FLOW_FINISH_GATE: ok`. Its first line was `Resuming failed run finish-fb7a400f from step 3`.

Steps 1 and 2 — `lint` and all **4,137 tests** — did not re-run. The `ok` covered a tree that predated the fix, and **the tests had never executed against the file that had just changed.**

## Why this is a footgun rather than an edge case

The loop is: gate fails → fix the failing step → re-run → green. That is the natural workflow, and it is wrong **exactly when the fix touches code** — which a lint or typecheck repair almost always does. The step that would exercise the edit is the one the resume skips. *The more ordinary the workflow, the more likely the false green.*

The marker is not lying. It answers *"did the steps I ran pass"*, and the reader supplies *"and therefore this tree passes"*.

## The distinction the runner does not make

Resume is **correct for a crash** — an interrupted run left the tree unchanged, so completed steps are still valid. Resume is **wrong for a repair** — the tree changed, and the already-passed steps are precisely the ones whose verdict the change invalidates.

`runner.py:170` documents the intent as "crash-safe resume", which is sound. The runner simply cannot distinguish the two cases, and does not ask.

## Detection, and why the evidence disappears

Check the gate's **first** lines for `Resuming failed run`. If present, the `ok` does not cover steps before the resume point.

State lives at `<project_root>/.claude/runs/<run_id>.json` and is **consumed on success** — so the moment a resumed run goes green, the evidence that it was partial is deleted and the next invocation is fresh. A reader inspecting afterwards finds nothing.

## Options, not a design

- A `--fresh` flag, and/or `flow-finish-gate.sh` defaulting to it.
[...]

PR #1057, opened 2026-09-19
fix(gates): give every gate verdict its own exit code, and every stage its coverage (Closes #1027)

### LINKED ISSUE #1027: flow-finish-gate: all six warn verdicts exit 0, so the documented grading method cannot see a gate that did not run
Consolidated from the Nit Store (#864), 2026-09-16 sweep. Five comments, one reader.
All five re-verified against `843ba08`.

## The core: the documented grading method cannot see the verdicts that matter

`/flow:auto` Step 1 states the invocation discipline as *"Helpers print their own status
markers and exit codes carry the result."* That is true of `fail` and false of every `warn`.

Verified today in `scripts/flow-finish-gate.sh`: **all six `warn` verdicts `exit 0`**
(`:283`, `:295`, `:301`, `:545`, `:553`, `:559`). The exit code separates `fail` from
`{ok, warn}` and NEVER separates `ok` from `warn`. A caller following the documented rule
(`if gate; then proceed; fi`, `gate && git push`) waves through all six, including:

- `warn (carried, unverified: ...)` - added by #804 precisely to catch a run the runner
  could not verify the tree for. The one verdict that exists to stop an unverifiable carry
  is invisible to the grading method the docs prescribe.
- `warn (not run by fallback: ...)` and `warn (skipped gates: ...)` - both report that
  stages DID NOT EXECUTE, and exit 0 saying so.

https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5653379879

**The same shape one layer down, and in a container it is the NORMAL path:**

```
$ flow-finish-gate.sh --check-summary
NOTE: lib.cicd unavailable (CPP checkout not found); skipping Makefile completeness check.
FLOW_FINISH_GATE: skipped
rc=0
```

`skipped` and `passed` share an exit code. Measured from inside a Kyle session container,
where the degraded path is the ordinary state rather than an edge case. The same comment
records that CLAUDE.md's host-dependency count was stale (it said 8 of 52 scripts need the
checkout; the container measured 2 of 54) - a reader who trusts the count will not expect
the skip path to be the common one.
https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5647347734

## Three siblings that make the same green unreadable

- **`step_details` carries coverage evidence for `test` only**, so the standing "grade
  per-stage, not the rollup" procedure is unfalsifiable for three of four stages. `lint`,
  `typecheck` and `security_scan` emit `{id, status}` and nothing else. `status: "success"`
  is returned both by a stage that ran and found nothing and by one that no-opped. Only
  `test` carries `empty_invocations: 0`, which is exactly the "it ran and was not empty"
  assertion the other three lack.
  https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5654120412
- **`lib.security gate` emits only findings, never a verdict.** A PASSING run's complete
  output was 22 lines, every one `WARNING: HIGH: ...`. No summary line, no threshold, no
  counts. The verdict exists exclusively in `$?`, which is exactly what a pipe destroys -
  `python -m lib.security gate flow_finish | tail -20` renders a passing gate as an
  unbroken wall of HIGH warnings. A failing run would differ by one line out of 23.
  https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5653359789
- **`lib/cicd/runner.py:543` asserts cause on evidence that cannot establish it.** The
  rerun-pass log line reads `RE-RUN PASSED - first attempt was a flake`. What was observed
  is a pass from a re-run narrowed by `--last-failed` to the failed id alone - collection
  cut to 58 items, 57 deselected. That narrowing is precisely the treatment that turns an
  ORDER-DEPENDENT failure green, so the evidence and the claim point in opposite directions.
  The machine-readable `warn (rerun passed: ...)` is correct; the log line is what a human
  reads when deciding whether to chase the failure, and it says the question is settled.
  https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5654148028

## Shape of the remedy

Two halves, either of which closes the reader-facing hole alone:

1. Give `warn` a distinct exit status (3), keeping 0 = clean and 1 = fail, and document it,
   so wrappers grading on `$?` stop on a warn rather than proceeding. `skipped` likewise.
2. Or leave the status and FIX THE SKILL TEXT: state that this helper's result is the
   `FLOW_FINISH_GATE:` line, that only the literal token `ok` is a pass, and that `$?`
   distinguishes only `fail`. The current state is the worst of both, because the docs
   actively point at the channel that cannot carry the distinction.

For `lib.security gate` and the per-stage counts, one terminal line on EVERY exit path
naming the gate, the verdict, the threshold applied and the counts measured against it -
so the passing and failing renderings differ by a line that SAYS which happened rather
than by a line that has to be counted.

## Negative control (ADR 0008)

`flow-finish-gate.sh` is a finish gate: it lets work through and nothing downstream
re-derives its verdict, so it is squarely inside the bound.

- Red case for the exit-code half: a run that produces `warn (skipped gates: ...)` must
  make `if gate; then echo PROCEED; fi` NOT print PROCEED. Today it does.
- Red case for the per-stage half: a stage configured to match zero files must report a
  zero-coverage marker rather than `status: "success"`. A clean stage must not.
- Red case for the runner wording: a test that fails only under a specific sibling order
  must NOT be reported as "a flake" by a re-run that deselected the sibling.

cooneycw:
## What was wrong

#1027 reports one defect in two layers.

**The verdict could not be read from the channel the docs point at.** `/flow:auto` states the rule as *"helpers print their own status markers and exit codes carry the result."* That was true of `fail` and false of every `warn`: all six `warn` verdicts exited 0, so `$?` separated `fail` from `{ok, warn}` and never `ok` from `warn`. A caller following the documented `if gate; then proceed; fi` shape waved through all six — including `warn (carried, unverified: ...)`, added by #804 precisely to stop a carry the runner could not verify, and `warn (skipped gates: ...)`, which reports that stages did not execute. `skipped` shared the same 0, so in a container — where the degraded path is the ordinary state — `skipped` and `passed` were indistinguishable.

**Three of four stages could not say whether they examined anything.** `step_details` carried coverage evidence for `test` alone, so the standing "grade per-stage, not the rollup" procedure was unfalsifiable for `lint`, `typecheck` and `security_scan`: `status: "success"` was returned both by a stage that examined the tree and by one that no-opped.

## What changed

**Exit codes** — `ok` 0, `fail` 1, usage error 2, `warn` 3, `skipped` 4. `warn` and `skipped` get *different* codes: "ran, passed, something is soft-bad" and "did not run, no evidence" are facts a caller must branch on separately. 2 is left alone because it already means "usage error" in this script.

**Per-stage coverage** — `lib/cicd/coverage.py` parses what each tool says about its own coverage into three states (`covered` / `zero` / `unknown`), carried on `StepResult`, in `step_details`, and in a top-level `coverage` object the gate greps. Only `zero` warns. `unknown` is recorded but never graded on — the same bound `runner.py` already draws for an unparseable test summary, because warning on silence fires on every harness CPP cannot parse, on every run.

**Reachability is stated, not implied** — and this is the part worth reading:

| stage | coverage | why |
|---|---|---|
| `typecheck` | real count, both verdicts | mypy states `checked N source files` |
| `security_scan` | real count, **component-scoped** | the gate line now carries `secrets-scanned=`, named for the check it measures |
| `lint` | **`unknown` in practice** | ruff states no positive count, and its "No Python files found" warning is **suppressed whenever a `pyproject.toml` exists** |

That last row was measured across four tree shapes (with and without a `[tool.ruff]` section). Every real Python project has a `pyproject.toml`, so the ruff detector is blind in exactly the population the gate runs against. It was found by running the control — the parser was correct and the gate was green. The pattern is kept (correct when it appears), `lint` reports `unknown`, the limitation is written at the code, and a real fix (measuring the stage's *input* rather than parsing its output) is filed at #864 rather than half-built here.

The **fallback lane** detects the same markers directly: in a container that lane is the ordinary path, so a runner-only fix would have left the measured case blind. Its non-test gates now `tee` and grade on `PIPESTATUS[0]`, never the tee's status.

## Two parts of the issue needed no code

- The runner's `RE-RUN PASSED - first attempt was a flake` wording was removed by `f6c6071` (#769) and reinforced by #900 — closed 2026-09-13, **three days before #1027 was filed**. `git show 843ba08:lib/cicd/runner.py`, the issue's own stated verification base, contains zero occurrences, so that re-verification did not hold for this item.
- CLAUDE.md's stale "8 of 52 scripts" count does not reproduce: the file is 164 lines and carries no such claim.

## Negative control (ADR 0008)

`flow-finish-gate.sh` is a finish gate — it lets work through and nothing downstream re-derives its verdict — so it is squarely inside the bound.

| | input | verdict | exit |
|---|---|---|---|
| **RED** | zero-coverage tree, `origin/main` 967c098 | `ok` | **0** |
| **GREEN** | *same tree*, this branch | `warn (zero coverage: security_scan)` | **3** |
| **POSITIVE** | tree with one real source file | `ok` | **0** |
| **POSITIVE** | CPP's own tree | `ok` | **0** |

The last row is the one that proves the instrument is not blind: typecheck reports **243 source files** and the secrets scan **590** on this repo — a green from a working instrument, not an absent one.

Detector mutations, each restored byte-identical afterwards: blind ruff parse → 3 tests red; always-`ZERO` aggregate → 3 red; shell awk that never matches → 1 red; reverted resume rebuild → 2 red; reverted identity attribution → 1 red; global dedupe restored → 1 red.

## Counter-model review

Reviewer: **codex/gpt-6-astra** (must not be the implementing model), 2 passes
Findings: **8 accepted (fixed)**, 0 rejected, 0 deferred
Red cases proposed: 8, already covered by our tests: 7 — the eighth (the fallback-lane detector) was closed by three new tests
Receipt: `docs/measurements/counter-model/2026-09-19T124017Z-issue-1027.json`

All eight were real, and four of them were the same class of defect as the issue itself:

1. **Resume dropped carried coverage** from the roll-up the shell gate reads — #1027's own false green, restored by resume, for the stage #1027 is about.
2. **`unknown` was omitted rather than recorded**, contradicting the docstring — and a test asserted the omission, pinning the defect.
3. **Unreadable files counted as examined** (`files_scanned` incremented above the `try`), so one unreadable file reported `secrets-scanned=1` for a scan that opened nothing.
4. **A component's count was read as the stage's** — `secrets-scanned=0` claimed "this gate proved nothing" about gitignore and permissions checks that had examined their subjects. Fixed by carrying a `scope` with the number.
5. **A mixed aggregate attributed a neighbour's files to the named component** ("the secrets scan examined 47 source files" — mypy's 47).
6. **A global dedupe collapsed distinct within-stream invocations**, so a step's measured coverage changed because an unrelated stream had output.
7. **Consumer instructions still carried the stage-wide overclaim.**
8. **Resume dropped the partially-empty carry** (`covered` with `empty_invocations > 0`).

Each carries a test with a demonstrated red.

## Test plan

- `make lint` + `make test` + `make typecheck` + security scan via `flow-finish-gate.sh` → **`FLOW_FINISH_GATE: ok`**, 4169 passed
- `make scripts-inventory-check`, `make instrument-census-check`, `check-negative-controls.py` → all ok
- 34 new tests across `tests/test_cicd_coverage.py` (new), `test_runner.py`, `test_flow_finish_gate.py`, `test_security_scanners.py`

Closes #1027


cooneycw:
Detail on the parked sibling (per cpp-o1's request to record my sizing judgement here for whoever resumes this):

"`step_details` carries coverage evidence for `test` only" - lint/typecheck/security_scan entries in the runner's JSON are bare `{id, status}`, so a no-op stage and a real pass both read `"status": "success"`. This is NOT one uniform patch - each tool needs its own parser, and they differ a lot in difficulty:

- **mypy** - cheap. It already prints `Success: no issues found in N source files` on its own success line; a coverage count is a straightforward regex against existing output.
- **security_scan** - now cheap, now that this PR's sibling fix landed: `cmd_gate` prints `SECURITY_GATE: <gate> PASS|FAIL (blocked=N warned=M; ...)` unconditionally, so wiring the runner to capture that line into `step_details` is comparatively small - the hard part (emitting a parseable count) is already done.
- **ruff** - NOT cheap. `ruff check .` prints nothing about file count on a clean pass by default (`All checks passed!` only). Getting a real count needs either changing the invocation itself (`--statistics` or similar - a `lib/cicd/steps.py` step-definition change, not just a parser) or a separate file-count probe. Both `flow-finish-gate.sh`'s fallback and `lib.cicd`'s step definition invoke `ruff check .` directly today.

So whoever picks this up has three sub-decisions, not one: do the two cheap wins (mypy, security_scan) and treat ruff as a residual, do all three now, or decide this belongs in its own issue rather than #1027 (my lean, stated to cpp-o1 in the message that led to this pause - the coverage gap is a different shape of problem from the exit-code confusion the rest of this PR fixes, and deserves its own negative controls and review rather than riding in on this diff).

Issue #1056, opened 2026-09-19
git stash's shared stack bit again through /cpp:update and ad-hoc use — the flow lane's fix is a convention in two documents, so it cannot reach either


cooneycw:
## A known, documented, already-fixed hazard recurred through the one path the fix does not cover

`/flow:auto` and `/flow:finish` were changed to commit-first precisely because of this, and both carry the reasoning in-file:

`.claude/commands/flow/auto.md:655-663`:

> Stashes live in the repo's COMMON git dir, so every linked worktree shares ONE stack: with ~7 concurrent sessions, two runs stash-pushed seconds apart and each bare `git stash pop` silently restored the OTHER session's uncommitted work into the wrong worktree. A WIP commit is branch-local — a sibling session cannot take it […] **NEVER use `git stash push` or a bare `git stash pop` in a shared-worktree flow.**

`.claude/commands/flow/finish.md:48-51` says the same. **So the lifecycle is safe and this issue is not a claim that it is not.**

**What is still exposed is `/cpp:update`.** `.claude/commands/cpp/update.md:101`:

```bash
git stash push -m "cpp-update auto-stash $(date +%Y%m%d-%H%M%S)"
```

That runs in the main checkout, on a stack shared with every linked worktree, and it is the one documented flow that still pushes.

## It recurred today, measured

A fleet worker running `/codex:auto` on #1032 ran `git stash push -- scripts/worktree-remove.sh` in its **own** worktree, ran a test, then `git stash pop`. The pop restored `scripts/flow-finish-gate.sh` — 67 lines of a **different session's** in-progress #1027 work — and the worker's own stash entry was gone from the reflog.

Nothing was lost, because the worker handled it correctly: it did not discard the unexpected diff, recovered its own commit via `git fsck --unreachable --no-reflogs` by matching the stash's commit message, and re-stashed the stranger's work under a label rather than dropping it.

**The mechanism, verified structurally rather than inferred from the symptom.** Three sibling worktrees:

```
issue-1027:  git-dir        .git/worktrees/claude-power-pack-issue-1027
             git-common-dir .git                      <- shared
issue-1032:  git-dir        .git/worktrees/claude-power-pack-issue-1032
             git-common-dir .git                      <- shared
issue-1046:  git-dir        .git/worktrees/claude-power-pack-issue-1046
             git-common-dir .git                      <- shared
```

[...]

Issue #1032, opened 2026-09-16
Worktree ownership records cannot distinguish held from abandoned: an ownerless lock, a directory name that lies, a recycled pid, and a live session inside a removed tree


cooneycw:
Consolidated from the Nit Store (#864), 2026-09-16 sweep. Eight comments. Every one is a
record read as evidence of a live owner by something deciding whether removal is safe.

## The class

`git worktree list` prints `locked` identically whatever the lock says. A claim file, a
directory name and a pid are each treated as identity, and none of them is. The two error
directions are both live here: a tree that cannot be reclaimed, and a tree that a sweep would
delete out from under open work.

## Specimens

**1. A lock with no owner reads as "someone owns this".** Observed on
`claude-power-pack-issue-921-preflight-freshness-window`: `.git/worktrees/<name>/locked` held
the literal `added with --lock` - git's default string. No issue, no pid, no session, no
timestamp. The other three trees that day held a real
`flow-claim issue=1012 pid=4093123 session=e025b3f5-... host=... ts=...`.
`worktree-remove.sh` reported:

```
Claim:    foreign (issue #-, pid -, session -)
Another /flow session is driving this checkout. Removing it would destroy
its uncommitted work - the failure this claim exists to prevent (issue #597).
exit 4
```

Every field empty, and the surrounding facts contradicted all three assertions: no process
had a cwd inside, `git status --porcelain` was 0 lines, and its PR was MERGED. **The refusal
direction is right** - an unidentifiable claim must not read as free - but
`foreign (issue #-, pid -, session -)` and `foreign (issue #921, pid 12345, session abc)` need
opposite remedies and print the same sentence. The consequence is that such a tree is
permanently unremovable except via `--steal`, so the safe path and the deliberate override
collapse into one, which is how `--steal` stops meaning what it says.
Note the failure is two-sided: nothing records WHO locked it, so had that tree held unpushed
work there would be no session to ask.
https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5695560311
https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5696764169

**2. A worktree directory name can hold a DIFFERENT issue's branch.** Observed:
`.../claude-power-pack-issue-971-workers-cannot-escalate-an-unacknowledged-message` had branch
`issue-980-lexicon-quoted-token` checked out. The directory says #971 (CLOSED, PR #1004
[...]

Issue #1023, opened 2026-09-16
flow-driver-capability list reports cannot=- for every driver: #877's meta axis never reached the list verb's two call sites


cooneycw:
Measured on `f1c82d8` (worktree for #1017).

## The matrix reports that no driver has any incapacity

`scripts/flow-driver-capability.sh list` is what `/flow:auto`'s own capability-contract
section tells an orchestrator to run ("the whole matrix"). It currently answers:

```
$ ~/.claude/scripts/flow-driver-capability.sh list
scripts/flow-driver-capability.sh: line 291: $4: unbound variable
  flow:auto    general              web=yes  container=yes  cannot=-
scripts/flow-driver-capability.sh: line 291: $4: unbound variable
  codex:auto   implementation-only  web=no   container=no   cannot=-
scripts/flow-driver-capability.sh: line 291: $4: unbound variable
  qwen:auto    implementation-only  web=no   container=yes  cannot=-
scripts/flow-driver-capability.sh: line 291: $4: unbound variable
  gemma:auto   implementation-only  web=no   container=no   cannot=-
FLOW_DRIVER: known
```

`cannot=-` for every driver, including three that are implementation-only,
web-less, and meta-forbidden. `show` on the same driver is correct:

```
$ ~/.claude/scripts/flow-driver-capability.sh show codex:auto | grep CANNOT
FLOW_DRIVER_CANNOT=research web container meta
```

Exit code is 0 and the marker line still reads `FLOW_DRIVER: known`, so nothing
about the verdict says the row is empty because the function aborted.

## Cause

`#877` added the `meta` axis: `driver_record` grew two fields and `fence_of` /
`cannot_of` grew a 4th parameter (`scripts/flow-driver-capability.sh:274,291`).
The `show` (`:316`) and `check` (`:515`) call sites were updated. **Both call
sites inside the `list` verb were not**, and neither unpacks `meta` out of the
record at all:

- `:400-414` (`list --json`) - unpacks 6 of 8 `|`-delimited fields, then calls
  `fence_of "$scope" "$web" "$container"` and `cannot_of "$scope" "$web" "$container"`
- `:425-432` (`list` plain) - same, three-argument `cannot_of`

Under `set -u` the missing `$4` aborts the function, so it returns the empty
string; `${cannot:--}` then renders `-`, which is the helper's documented word
for "there is none".

## The `--json` form is wrong in a third way

The under-split record leaks `meta` and `meta_basis` into `container_basis`:

```
[...]

Issue #555, opened 2026-07-07
codex-skill-gen: emit Codex SKILL.md skills from .claude/commands SoT


cooneycw:
Companion to codex-power-pack epic cooneycw/codex-power-pack#64 (plugin-marketplace-modernization, ratified hybrid SoT).

Evolve `scripts/codex-prompt-sync.py` (flat custom prompts) into a Codex SKILL.md generator:

- [ ] Per-command skill dirs: `<family>-<cmd>/SKILL.md` with frontmatter name/description; progressive-disclosure layout; bundle referenced helper scripts
- [ ] Harness transforms: strip/translate Claude-only constructs (Agent tool, native worktrees, `!` prefix, /plugin refs); front-load trigger words in descriptions (Codex truncates long skill lists)
- [ ] Same ownership rules as today: GENERATED markers only, hand-curated files untouched, git-free --check for the validate container

**Acceptance:** generator emits valid Codex skills for all surviving families; drift --check wired into CPP CI; codex/prompts/ flat surface marked deprecated pending cutover.

PR #1058, opened 2026-09-19
fix(flow): distinguish an unparseable lock from a live claim, fix a boundary match, refuse clean occupancy (Refs #1032)

cooneycw:
## Summary

Chunk 1 of 3 for #1032 (worktree ownership records cannot distinguish held
from abandoned). Covers specimens **#1, #5, #6, #7, and #4a**, all localized
to `scripts/worktree-remove.sh`. Specimens #2 (identify by branch, not
directory name - spans several other flow helpers not yet enumerated) and
#4b (a session noticing its own cwd disappeared) are explicitly out of scope
here. #4b has been filed in the *kyle* repository as
[kyle#1225](https://github.com/cooneycw/kyle/issues/1225) rather than handled
here: nothing in claude-power-pack observes a running session - Kyle has the
reconciler and process-state/readiness split this needs. The split is
deliberate: **#4a is prevention** (removal declines - CPP owns removal, so
CPP can decline), **#4b is detection after the fact** (a different control
plane's job).

## Changes

**Specimen #1** - `flow-worktree-claim.sh check` returns `STATE=foreign` when
a lock's reason text does not parse as this project's own
`flow-claim issue=... pid=...` format (e.g. git's own default "added with
--lock" when a worktree is locked without `--reason`). In that case every
owner field is genuinely empty - there was nothing to parse. `worktree-
remove.sh` grouped `held | foreign` into one case arm asserting "Another
/flow session is driving this checkout" even when every field read `-`, so an
unidentifiable lock read as an identified one and was permanently
unremovable except via `--steal`. Split the arms: `held` (a real, parseable,
live claim) keeps its exact hard-stop behavior; `foreign` now prints an
honest message and falls through to the independent occupancy/dirty/unpushed
checks - the same way `free`/`unsupported`/`unknown` already do - with a
`git worktree unlock` right before the final removal, reached only once every
check above has already decided removal is safe.

**Specimen #5** - `INSIDE_WORKTREE` used a bare `"$CWD" == "$WORKTREE_PATH"*`
prefix match, so a shell whose cwd is a sibling directory that merely starts
with the same string (e.g. removing `/foo/x` while standing in `/foo/x-2`)
was wrongly judged to be inside the target. Now uses the same
exact-match-or-genuine-child pattern `occupancy_scan()` already uses
correctly for the identical class of bug.

**Specimen #4a** (folded into this chunk during review) - occupied-but-clean
silently proceeded with removal ("nothing unrecoverable to lose, so removal
stands"). **This is a behavior change, not merely a bug fix** - flag it in
release notes: a clean snapshot right now says nothing about what a live
process writes a moment later into what becomes an unlinked inode after
removal, which is exactly the incident #1032 describes (a live session
running inside a REMOVED worktree, writing to an unlinked inode), and was
reinforced empirically during this round: 10 live processes were found in a
worktree while every other available signal called it safe to merge.
Occupied-clean is now a refusal (exit 5, the same code already used for
occupied-dirty - both share the same risk factor: a live process has this
worktree as its cwd). `--force` still does not suppress it; `--steal` remains
the override. This inverts `test_occupied_but_clean_is_still_removed`
(renamed `test_occupied_and_clean_is_now_refused`), whose docstring
explicitly defended the old behavior - the historical reasoning is kept as a
code comment, following the identical precedent issue #899 already
established in this file for an equivalent inversion.

**Specimens #6/#7** - the header Exit codes table and both `--help` usage
synopses were stale relative to issue #899 (missing exit 6/7 entirely,
`--steal`/`--allow-dirty`/`--allow-unpushed` missing from both usage lines
despite being documented in the option list immediately below). Resynced to
actual behavior, including exit 5's broadened trigger from this PR.

**Also resyncs 6 bundled copies** of this script under
`codex/skills/flow-{auto,cleanup,doctor,merge,start,wave}/scripts/
worktree-remove.sh` via `codex-skill-sync.py --write flow` (issue #555) -
deterministic, git-free regeneration, unrelated to the logic change. The
source edit left them stale; `codex-skill-sync.py --check` caught it - the
same mechanical gap #1023 hit for a different script, and the second time
today the bundled mirrors have drifted under a source edit with only
`--check` noticing. 8 files in the diff for a 2-file logic change is this
reason, not scope creep.

## Negative controls (ADR 0008)

Confirmed RED **twice** against the actual rebased branch base (`13b6734`),
not carried forward from an earlier check - once before delegating
implementation, once again after the final rebase, using a plain
`git show <base>:path > file` swap rather than `git stash` (issue #1056: this
repository's shared `.git` directory means `git stash` is unsafe across
sibling worktrees of one repo - do not use it here):

- `test_native_foreign_lock_is_checked_then_safely_removed` - a real
  `git worktree add --lock` lock, clean tree, no process inside - failed
  with the exact false-claim message on pre-fix code, passes after.
- `test_invocation_from_sibling_path_prefix_is_not_inside_target` - failed
  by wrongly printing "Currently inside worktree being removed" on pre-fix
  code, passes after.
- `test_occupied_and_clean_is_now_refused` - failed by removing the
  worktree instead of refusing on pre-fix code, passes after.

**GREEN control**: `test_live_parseable_flow_claim_is_still_refused` - a
genuinely live, parseable claim, naming a real running pid, must still
refuse with real field values on both the pre-fix and post-fix script.
Without this, "foreign now falls through" could be satisfied by a fix that
simply stopped refusing anything, and every RED test above would still pass.

## Cross-model review

Implementation delegated to Codex CLI under `--sandbox workspace-write`
(model: `gpt-5.6-sol`, verified from this run's own rollout thread, not the
newest rollout on this host - this host had at least one other session
writing fabricated model names to its own rollouts during this window).
Codex's own test run hit a sandboxed network failure trying to install a
transitive dependency; it recovered using a locally cached pytest
environment and reported 137 tests passing. Independently re-verified in a
networked environment rather than trusting that report alone: same 137/137
result, both immediately after delegation and again after the final rebase.

Reviewed by Claude Code: diff matches the isolated defects and named line
ranges exactly. The occupied-clean behavior change (specimen #4a) was stated
as an explicit `AUTHORIZED CHANGES` item in the delegation prompt rather than
left to the delegate's own judgment, since it inverts a previously-deliberate
test assertion. No scope creep, no files touched outside
`scripts/worktree-remove.sh`, the three existing `tests/test_worktree_
remove_*.py` files, and the 6 generated bundle mirrors. No unexpected
commits or pushes from the delegated run.

## Test plan

- [x] `uv run --extra dev pytest tests/test_worktree_remove_occupied.py tests/test_worktree_remove_refusal.py tests/test_worktree_remove_squash.py -v` - 137/137 passed
- [x] `python3 scripts/codex-skill-sync.py --check` - 74/74 skills in sync
- [x] `uv run --project ~/Projects/claude-power-pack python -m lib.cicd run --plan finish` - 4/4 steps success (lint, test: 4084 passed, typecheck, security_scan), read from the JSON `"success"` marker, not the wrapper's exit code
- [x] Negative controls confirmed RED on pre-fix code, GREEN control confirmed unaffected on both

Refs #1032


cooneycw:
**Recording why the occupancy verdict has four states and not two**, because the code gets this right and the reason is not written down anywhere the next reader will find it.

`worktree-remove.sh` now emits `WORKTREE_REMOVE_OCCUPANCY:` as one of `clear` / `unknown` / `occupied-clean` / `occupied-dirty`. Two of those distinctions are load-bearing and both are easy to collapse by accident:

**`clear` is a measurement, not an ownership claim.** A `/proc` scan that finds processes is decisive for *in use*. An empty scan is **not** proof of no owner — a session paused between tool calls has no process running at all, so "scanned, found nothing" and "nothing is there" are different facts. The word `clear` says the first. Anything that renames it to `free`, or reports it as "this tree is safe to remove", asserts the second and re-creates specimen #1 in a new place: a verdict that sounds like knowledge and is actually an absence of evidence.

**`unknown` is not `clear`.** Could-not-look and looked-and-found-nothing are distinct, and folding them loses the only signal that the instrument failed.

Concrete case from this host today, which is what the specimen #4a change is defending against: a PR was nearly merged as an abandoned inherited PR — `MERGEABLE`, `BEHIND`, green on both lanes, three days old, no review requested. Every signal available through the API said stale and safe. A `/proc` scan of its worktree found **10 live processes**, parented to an interactive session five minutes into its own `/flow:finish`. The tree was in active use and nothing except the cwd scan could tell.

Had it been swept on an "occupied but clean" reading, that session would have lost uncommitted work and nothing it could see would have warned it. That is the empirical half of why the test docstring's premise — *"with nothing unsaved, nothing is lost"* — is false rather than merely outdated: a clean snapshot now says nothing about what a live process writes a moment later into what becomes an unlinked inode.

A negative control for any future change here needs all four states produced, not two — and the `unknown` one is the awkward one to fabricate, which makes it the one most likely to be skipped.

— cpp-o1 (orchestrator, `cpp-issues` fleet run)

Issue #1048, opened 2026-09-19
Every counter-model receipt ever written records the same reviewer, so the field cannot distinguish the models it exists to distinguish


cooneycw:
## The measurement

`--reviewer` exists to answer one question, the one ADR 0007 turns on: **which model reviewed this, and was it a different model from the one that implemented it?**

A read-only sweep of `docs/measurements/counter-model/` (run 2026-09-19 by a peer session, nothing modified) returns:

```
13  codex/gpt-5.5
```

**13 of 13. No other value has ever appeared.** A field that takes one value on every run is not recording anything - it is echoing a constant. That is the signature of a copied literal rather than of thirteen independent observations, and it means the field cannot currently answer the question it was built for.

The extractor is not blind: the sweeping session found its own receipt, whose value it already knew, so a zero would have meant a broken sweep rather than a clean tree.

## What is verified, and what is not

This boundary is load-bearing - please do not let it collapse into "13 known-bad receipts", which is **not** what was measured:

- **Verified wrong: 1 of 13.** `2026-09-16T205221Z-issue-1017.json` records `codex/gpt-5.5`. The session that produced it still had the `codex exec` transcripts and both passes read `model: gpt-6-astra` / `reasoning effort: high`. Both predate the 2026-09-19T08:25:07Z config change, so the `gpt-5.6-sol` switch is not the cause.
- **Unverified: 12 of 13.** Their transcripts were not inspected. Some date from 2026-09-15 and may genuinely have run on `gpt-5.5`. Nobody has established that they are wrong, and nobody should assert it.

## Cause

`.claude/commands/flow/auto.md` carried `--reviewer "codex/gpt-5.5"` as a literal in its worked example. Runs copied it. #1045 removes the literal and replaces it with a thread-anchored lookup, which stops new receipts from inheriting it - but #1045 is forward-only and does not touch what already landed.

The literal being *documented* is what makes this worse than a typo: it reproduced faithfully on every run that followed the instructions correctly.

## The durable fix, and why the current shape invites this

`--reviewer` is caller-supplied. That makes the reviewer's identity **a marker written by the thing being measured** - which is precisely the failure `docs/scripts.md` already warns about for the PR-body block, reproduced one level down inside the receipt itself. PR #1000's lesson was that a hand-written marker is defeated by whoever writes it; the receipt was the remedy, and it has the same hole.

Proposed shape, for discussion rather than as a settled design:

[...]

Issue #1047, opened 2026-09-19
counter-model receipt hardcodes --implementer "claude/opus-5", so the one property the receipt exists to check is asserted, not measured


cooneycw:
## The claim that is never checked

ADR 0007's requirement is a property: **the reviewing model must not be the implementing model.** The counter-model receipt is the artifact that records whether it held.

`/flow:auto` Step 6 item 1d writes:

```
--reviewer "codex/<model>" --implementer "claude/opus-5"
```

#1045 makes the reviewer half a measurement, read from the review's own thread rollout. The implementer half remains a **literal**, copied from the example on every run.

## Failure scenario

A receipt's two identity fields are the only evidence that two different models were involved. One of them is now measured and the other is asserted, so the equality check they exist to support cannot fail for the asserted side:

- A run driven by `/codex:auto`, `/qwen:auto` or `/gemma:auto` - where Codex, Qwen or Gemma implements - still writes `claude/opus-5`.
- A different Claude model implementing writes `claude/opus-5`.
- Most sharply: if the **same** model both implemented and reviewed, the receipt records two different names and the diversity check passes. That is the exact condition ADR 0007 forbids, recorded as satisfied.

The value is also stale in the ordinary way - it names a specific model version that will move - but staleness is the smaller problem. The structural problem is that a field nothing derives cannot report the state it exists to detect.

## Why this is its own issue

#1045 fixes the reviewer side and stops there deliberately. Measuring the implementer is a different problem with a different mechanism: a session does not have a rollout record the way a `codex exec` invocation does, so the answer is not "do what #1045 did to the other field". It needs a decision about where an implementing session's identity comes from and what happens when it cannot be established.

## Scope

- [ ] Decide the source of the implementing model's identity, and what "unknown" means at that site
- [ ] Fail closed: write no receipt rather than a receipt with an assumed identity, in line with #1045's rule that an unknown reviewer is never a name
- [ ] Add the check the fields exist for - reviewer == implementer must be a REJECTED receipt, not a silently written one
- [ ] Negative control: a fixture where both identities are the same model, asserting the writer refuses it. Without that case the check is untested in the only direction that matters

## Found by

Counter-model review (codex, `gpt-5.6-sol`) of the #1045 branch, second pass.

Refs #1045, #934, #1001, ADR 0007

Issue #1046, opened 2026-09-19
counter-model receipt: a skip has no reviewer, but the writer requires one - so the documented skip path must fabricate or fail


cooneycw:
## The contradiction

`/flow:auto` Step 6 item 1d requires a receipt on **every** outcome, including a skip:

> **1d. Write the receipt - ALWAYS, including on a skip**

And a skip means no review happened. On `codex-absent` the binary was never run; on `reviewer-unavailable` it ran and returned nothing usable. In neither case does a reviewing model exist.

But `scripts/counter-model-receipt.py` requires `--reviewer` for every status and rejects both an omitted and an empty value. So the documented path has only two endings, and both are wrong:

- **fabricate** - name a model that did not review, which is exactly the false provenance #1045 is fixing on the `ran` path; or
- **fail** - write no receipt, which makes a skip indistinguishable from a stage that was never wired in, the precise failure 1d exists to prevent.

## Why it matters

The receipts are the evidence base for whether cross-model review happens at all (#934/#1001). A skip receipt naming a reviewer is worse than no receipt: it enters the measurement as a review that occurred. And because the reviewer field is also what a reader would use to check the ADR 0007 property - the reviewing model must not be the implementing model - a fabricated value can make that check pass on a run where nothing reviewed anything.

## Found by

Counter-model review (codex, `gpt-5.6-sol`) of the #1045 branch, second pass. It is a pre-existing defect that #1045 surfaced rather than introduced: #1045 makes the `ran` path measure its reviewer honestly, which is what makes the skip path's inability to say "none" visible.

## Scope

- [ ] Decide the representation: reviewer absent/null for skipped runs, or a distinct `attempted-reviewer` field that is explicitly not a claim that a review occurred
- [ ] Change the writer, the validator, and the schema together - a validator that still demands the field would reject every honest skip
- [ ] Commit executable skip examples for both `codex-absent` and `reviewer-unavailable`
- [ ] Negative control: a fixture that a skip receipt naming a reviewer is REJECTED. Without it the new tolerance silently accepts the fabrication it was written to forbid

## Deliberately not in scope

The `ran` path, which #1045 owns.

Refs #1045, #934, #1001, #1015

Issue #1045, opened 2026-09-19
codex command docs name gpt-5.5 as the reviewer model; the configured model is not that, and the docs cannot say so


cooneycw:
## The claim

`.claude/commands/codex/ask.md:94` states, as current fact:

> Defaults come from `~/.codex/config.toml` (currently `gpt-5.5`, `xhigh`).

It is wrong on both halves. Measured 2026-09-19: `~/.codex/config.toml` read `model = "gpt-6-astra"`, `model_reasoning_effort = "high"`. Neither `gpt-5.5` nor `xhigh` had been the configured value for some time, and nothing in the repository could tell a reader that.

Seven more forward-looking references repeat the stale name:

| File | Lines |
|---|---|
| `.claude/commands/codex/ask.md` | 2, 8, 75, 92, 94 |
| `.claude/commands/codex/code_review.md` | 8, 159 |
| `.claude/commands/flow/auto.md` | 881 (the receipt `--reviewer` example) |

## Why it matters beyond being untidy

`/flow:auto` Step 6 item 1d writes a counter-model receipt whose `--reviewer` value is copied from the example at `auto.md:881`. That example says `codex/gpt-5.5`. So a run that faithfully follows the document records, in a durable measurement, the name of a model that did not perform the review. The receipts are the evidence base for whether cross-model review is happening at all (#934/#1001) - a reviewer field that names the wrong model is worse than an absent one, because it looks like data.

This is #870's defect in its original form: a claim with no stamp, in a document a reader acts from.

## Not a blind find-replace

39 references to `gpt-5.5` exist repo-wide. **21 of them are historical records and must not be touched**: 12 counter-model receipts under `docs/measurements/counter-model/`, 2 CHANGELOG entries, `docs/research/forced-claim-prototype-2026-09-15.md`, `tests/fixtures/counter_model/README.md`, and provenance comments in `tests/test_lane_serveability.py` and `tests/test_negative_fixture_preconditions.py` crediting specific past findings to Codex. Each says what was true when it was written. Rewriting them would assert that `gpt-5.6-sol` reviewed work it never saw.

A further set is arbitrary fixture data rather than a claim about our model: the `--reviewer` arguments throughout `tests/test_counter_model_review.py`, the simulated reroute line in `tests/test_delegated_run_check.py`, and the parenthetical-spelling case in `tests/test_driver_capability.py`.

## Scope

[...]

PR #1059, opened 2026-09-19
feat(codex): derive the counter-model reviewer instead of accepting it (Refs #1048)

cooneycw:
**Resumed from a paused WIP state** (operator wind-down, previous session). Rebased onto `0c11b45`, re-verified end to end (all five negative controls were already complete before the pause - independently reran the pre-fix comparison again after rebasing), full quality gate green, and taken out of draft.

## Summary

A read-only sweep found ALL 13 landed counter-model receipts record the identical reviewer `codex/gpt-5.5` - the signature of a copied literal, not of 13 independent observations. `--reviewer` was a free-text CLI argument: a marker written by the thing being measured, defeated by whoever writes it (the same lesson a different marker - a PR-body heading - already taught this codebase once, per `scripts.md`).

## The fix

`--reviewer` is removed entirely as a flag for `--status ran`. Replaced with `--reviewer-exec-log PATH`: the writer derives the reviewing model itself, straight from the codex `--json` exec stream, rather than accepting a caller's claim about it:

1. Read the exec log's own `thread_id` (from its `thread.started` event).
2. Find the session rollout matching that thread_id under `--codex-sessions-dir` (default `${CODEX_HOME:-~/.codex}/sessions`, searched recursively, matched on thread_id alone - **never** "the newest rollout on the host", which is the exact defect #1045 already fixed once in the bash version of this same lookup).
3. Read that rollout's own `model` field.

Any step failing **REFUSES to write** - non-zero exit, no receipt file - rather than falling back to accepting a caller-supplied string from anywhere, including `/codex:code_review`'s own `CODEX_REVIEW_MODEL` marker.

### Why the `CODEX_REVIEW_MODEL` marker was rejected as a source

`CODEX_REVIEW_MODEL` is itself a caller-supplied string, so deriving the reviewer from it relocates the defect one level up rather than fixing it. Reading straight from the exec log is a source the script can own without depending on another command's stdout text.

### The framing

`auto.md:878` already said *"`--reviewer` is a MEASUREMENT, read it from the run - do not copy the literal."* #1045/#1049 made that instruction correct. **#1048 makes it structural**: the flag through which a wrong value could arrive stops existing.

`build()` and `validate()` are **unchanged** - the stored receipt schema is identical (a string or `null` in `"reviewer"`); only how `write()` arrives at that value changes. `--status skipped` is unaffected (#1046's rule stands). `--implementer` is **untouched** - still a required literal string; that is #1047, reusing this same derive/refuse shape once it lands.

## Six existing tests converted - what each still pins

| Test | Still pins |
|---|---|
| `test_a_skip_naming_a_reviewer_is_REFUSED_at_the_write_path` | a skip must not carry review evidence of any kind - same refusal, new flag name |
| `test_the_PROPERTY_is_enforced_not_just_documented` | a derived reviewer equal to the implementer is refused (fixture derives to match the implementer arg, exercising the same-model check post-derivation) |
| `test_a_DIFFERENT_reviewer_is_accepted` | counts/red_cases plumbing survives a normal write |
| `test_coverage_cannot_exceed_what_was_proposed` | the ratio check is unaffected by where the reviewer value comes from |
| `test_an_EMPTY_model_identity_is_refused` | **narrowed** to the empty-implementer half only - an empty reviewer can no longer be typed in as free text, so that half is structurally impossible to construct; docstring says so explicitly. Superseded by a *stronger* claim below (derivation genuinely cannot produce a value, not just "a string happened to be blank") |
| `test_two_runs_in_the_same_second_both_SURVIVE` | exclusive-create race safety survives a shared derivation fixture across two concurrent writes |

## New tests - the issue's required negative controls, plus the success path

- `test_reviewer_is_derived_from_the_matching_rollout` - a well-formed fixture derives `reviewer == "codex/<fixture model>"`.
- `test_a_missing_or_unreadable_exec_log_is_refused` (×2: nonexistent path, directory masquerading as a file).
- `test_an_exec_log_without_a_thread_id_is_refused`.
- `test_an_exec_log_without_a_matching_rollout_is_refused`.
- `test_a_matching_rollout_without_a_model_is_refused`.
- `test_thread_match_wins_over_a_newer_unrelated_rollout` - an unrelated rollout with a **later mtime AND a later-sorting filename** must not win over the thread-id-matching one. This is the case that would catch a future "just take the newest rollout" regression - the same trap #1045 fixed once already.

**RED, independently confirmed** against the pre-fix (post-#1046) script for all six new tests (seven cases counting the parametrized one) - not taken on the implementing model's word. Swapped in the pre-fix file via `git show HEAD:path > file` (never `git stash`, per #1056), reran: all seven fail with `unrecognized arguments: --reviewer-exec-log`, since those flags did not exist there at all. Restored the fix and reconfirmed 46/46 green - both before the wind-down pause and again after resuming and rebasing onto `0c11b45`.

## Also fixed: generated mirrors

This source edit touches two files that are themselves mirrored (`codex-skill-sync.py --write flow`): `codex/skills/flow-auto/reference.md` (documents this same call site) and `codex/skills/flow-auto/scripts/counter-model-receipt.py` (the bundled copy of the script). `--check` confirms all 74 skills in sync. **Five files in this diff for what reads as a two-file change is exactly that, not scope creep** - two files are hand-edited, three are mechanically generated from them.

## Deliberately NOT done here

- **`docs/measurements/counter-model/` (the 13 landed receipts) is untouched.** Their disposition - backfill where provable, mark unverifiable, or fix forward - is an **owner decision** per orchestrator guidance. Not made or suggested here. **1 of 13 is verified wrong** (the #1017 receipt, transcripts read `gpt-6-astra`); **12 of 13 are unverified**, not "known bad" - some may genuinely have run on `gpt-5.5`.
- **The corpus-uniformity guard** (issue scope item 5, "consider whether") is deliberately **deferred to its own issue, #1091**, not bolted on here: uniformity is not inherently wrong (13 identical values would be correct if one model genuinely performed all 13 reviews), so a guard that flags it needs a second source (the exec records) to tell "uniform because constant" apart from "uniform because true" - a different instrument, with its own membership floor and negative control. A gating version would also block on the same undecided 13-receipt disposition above.
- `.claude/commands/codex/code_review.md`'s `CODEX_REVIEW_MODEL` marker emission is untouched - unused by this new path, but removing it is a separate, independently-reviewable edit this change does not need.
- **No ADR 0008 census row added.** Verified with the orchestrator: this PR touches five files and none is a new instrument registration - `validate()`'s shape-only checking is caught by the surrounding pytest suite (the script's own docstring already explains why it carries no committed-control registration), so no row applies.

**#1048 stays open for scope items 1, 2 and 5** - the disposition of the 13 landed receipts, correcting the one verified-wrong receipt once that disposition is decided, and the uniformity guard now tracked separately as #1091. This PR closes none of them; `Refs #1048`, not `Closes`, so the issue's remaining state reads as intentional rather than forgotten.

## Test plan

- [x] `pytest tests/test_counter_model_review.py -v` - 46/46 passed
- [x] Same 6 new tests (7 cases) against the pre-fix script - all RED as described above
- [x] Full quality-gate run (`lib.cicd run --plan finish`): lint, tests, mypy, security quick scan - all green, both before the pause and again after rebasing onto `0c11b45`
- [x] `codex-skill-sync.py --check` - 74/74 skills in sync

Refs #1048


cooneycw:
## Adopted and finished by another session — acceptance accounting

The authoring session exited before merging. I gated this branch myself rather than trusting its author's green, and I am landing it as **partial delivery**: `Refs #1048`, not `Closes`. The issue stays open.

**Retitled from `WIP(codex):` to `feat(codex):`** — the PR title becomes the squash subject on `main`, and a commit reading `WIP` is not what this is.

### Against the issue's five scope items

| item | disposition |
|---|---|
| Make the writer derive rather than accept, and refuse rather than assume | **demonstrated** — `--reviewer-exec-log` derives via thread id → rollout → model; `--status ran` requires it; `--status skipped` refuses it |
| Negative control: derivation fails → writer REFUSES | **demonstrated** — five cases: missing/unreadable log, no `thread_id`, no matching rollout, rollout without a model, and `thread_match_wins_over_a_newer_unrelated_rollout` |
| Decide the disposition of the 13 landed receipts | **NOT DONE — owner's judgement**, as the issue states |
| Correct `2026-09-16T205221Z-issue-1017.json` | **NOT DONE** — history untouched, 0 receipt files changed; gated behind the disposition above |
| A guard that the reviewer field is not uniform | **NOT DONE** — the issue frames it as "consider whether", so it is not owed here |

### The premise moved; the defect did not

Filed on `13 of 13 receipts record codex/gpt-5.5`. Re-measured on `ac64cc1`: **13 `gpt-5.5`, 10 `gpt-6-astra`, 2 `gpt-5.6-sol`** across 25. The field discriminates now, so the symptom is gone.

That is not a reason to close it. #1045 fixed the *instruction*; this fixes the *mechanism*. The field discriminates today because agents are obeying a documented procedure — I am a live instance: my own receipt for #1031 reads `codex/gpt-6-astra` because I derived it by hand and typed it in. Correct, and correct only because I followed a rule. Until 2026-09-19 that document's worked example read `gpt-5.5`; a run copying it faithfully would have recorded a reviewer that never reviewed. Ten correct receipts prove ten agents were careful; they say nothing about the eleventh.

### Verification

- `FLOW_FINISH_GATE: ok` on the branch, run by me
- Woodpecker [pipeline 2073](https://woodpecker.essent-ai.com/repos/1/pipeline/2073) **success** on `083847a`
- Current with `main` (`ac64cc1`), 0 behind, clean tree

### What remains on #1048

Items 3 and 4 above — both owner decisions about the integrity of a merged measurement, deliberately left rather than guessed at.

Issue #1068, opened 2026-09-19
Codex consolidation: establish the migration spec, capability inventory and disposition ledger


cooneycw:
## Governing specification

Authoritative contract: [`.specify/specs/codex-consolidation/spec.md`](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/spec.md) (established by #1068). Governing sections: US1-US6, and the Evidence baseline.

Companions: [inventory](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/inventory.md) - what exists, at CPP `5ceb966a` / CxPP `681ea26b`; [ledger](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/ledger.md) - the disposition of every open obligation; [plan](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/plan.md) - phases and serialization; [review](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/review.md) - independent review, reviewer identity recorded.

Delivered by PR #1090 and left OPEN: acceptance item 3 is partial (this block IS the remainder) and item 6 is not discharged - Q1-Q9 are surfaced, not ruled on.

The spec is referenced, not copied: where it and the text below disagree, the spec is the later and authoritative statement, and the difference is a defect to fix in the spec.

---

Parent migration plan: #1067.

Status: proposed work, not authorization to implement. Before implementation, follow the approved governing spec established by the contract child and link its relevant sections. Existing issue acceptance is not superseded by this ticket.

Depends on: None; this is the first decision gate.

## Outcome

Turn the proposed migration in #1067 into an approved, authoritative contract before code moves. Eliminate ambiguity about what Codex users retain and what work is deliberately unfinished.

## Scope and acceptance

- Rebaseline both main SHAs; inventory native/shared/generated skills, helper/runtime modules, plugin families, invocation policy/evaluation assets, hooks/bootstrap/update/status flows, security/test protections, credentials/state, docs, installed sources and known downstream consumers. Do not expose credential values.
- Account for every open CxPP issue and PR, including PR cooneycw/codex-power-pack#239, and relevant CPP issues/Nit Store findings. For each capability or obligation record source, current delivery status, owner, destination and disposition: already covered, move, adapt, or owner-approved retirement. Identify unknown consumers as unknown, not absent.
[...]

Issue #1067, opened 2026-09-19
Consolidate Codex support in CPP and safely retire codex-power-pack


cooneycw:
## Governing specification

Authoritative contract: [`.specify/specs/codex-consolidation/spec.md`](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/spec.md) (established by #1068). Governing sections: the whole document.

Companions: [inventory](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/inventory.md) - what exists, at CPP `5ceb966a` / CxPP `681ea26b`; [ledger](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/ledger.md) - the disposition of every open obligation; [plan](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/plan.md) - phases and serialization; [review](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/review.md) - independent review, reviewer identity recorded.

**CORRECTED 2026-09-22 (orchestrator, claude-improvements wave).** This line read "Q1-Q9 are all open ... no child may resolve one on its own authority" and was STALE. Q1-Q9 were **resolved by the owner on 2026-09-20** - see [`spec.md` Open Questions](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/spec.md) ("Q1-Q9 were resolved by the owner on 2026-09-20") and [`ledger.md` §A](https://github.com/cooneycw/claude-power-pack/blob/main/.specify/specs/codex-consolidation/ledger.md), which carries all nine rulings with citations. Q10 was raised and answered by those rulings and **no longer blocks dormancy**. Reading this epic as the authority would have sent the owner back to re-decide nine settled questions; the spec is authoritative where the two disagree.

The spec is referenced, not copied: where it and the text below disagree, the spec is the later and authoritative statement, and the difference is a defect to fix in the spec.

---

## Outcome and recommendation

Make claude-power-pack (CPP) the maintained source for shared workflow, testing and security behavior, with explicit Claude and Codex adapters. Retire codex-power-pack (CxPP) only after Codex consumers work from CPP and an owner-approved rollback/cutover has been demonstrated. Keep useful Codex-specific capabilities; stop maintaining a second copy of the shared platform.

[...]

Issue #1086, opened 2026-09-19
ci: the pipeline is 84% one serial pytest step, and validate waits 30s for a lint it does not consume — measured 5.34x available


cooneycw:
CI numbers measured on `5ceb966` (pipelines 2039 / 2037 / 2036, all success on main).
Local suite numbers measured on `7b5ca75` (main fast-forwarded at 11:47:39, before the
first local run started; all four local runs are on that one commit, confirmed by reflog
showing no further move during them).

## The opportunity, measured

CPP's pipeline wall clock is ~290s, and one step is 84% of it.

| step | 2039 | 2037 | 2036 | on the critical path |
|---|---|---|---|---|
| clone | 3s | 4s | 3s | yes |
| secret-scan | 2s | 2s | 1s | yes |
| jq-stage | 3s | 2s | 1s | no |
| codex-skills-check | 3s | 3s | 1s | no |
| delegated-core-check | 3s | 2s | 1s | no |
| scripts-inventory-check | 2s | 2s | 1s | no |
| instrument-census-check | 3s | 3s | 1s | no |
| dependency-audit | 22s | 22s | 21s | no |
| eli5-vendor-check | 3s | 3s | 1s | no |
| eli5-upstream-drift | 3s | 3s | 2s | no |
| tool-risk-drift | 2s | 3s | 1s | no |
| **shellcheck** | **32s** | **32s** | **32s** | **yes** |
| negative-controls | 5s | 5s | 5s | no |
| **validate** | **250s** | **253s** | **262s** | **yes** |
| **wall** | **287s** | **291s** | **298s** | |

Reproduce:

```sh
woodpecker-cli pipeline ps cooneycw/claude-power-pack 2039 | sed 's/\x1b\[[0-9;]*m//g' | awk '
/^Step: /{n=$2} /^Started: /{s=$2} /^Stopped: /{e=$2}
/^State: /{if(n!=""){printf "%-26s %4ds\n", n, e-s; n=""}}'
```

Inside `validate` (step log, pipeline 2039): `uv sync` ~2s, `ruff check` <1s, **`pytest` 239.76s**, `mypy` ~7s.

The eleven fan-out steps already run concurrently. There is no step-level parallelism left to add. Both remaining costs are on the serial spine.

### 1. `validate` waits 30s for a lint it does not consume

[...]

Issue #1033, opened 2026-09-16
Supervision goes silently deaf four ways: an orphaned child holding the flock, a daemon that can never re-arm, pid-identity-blind liveness, and a documented latency the loop cannot provide


cooneycw:
Consolidated from the Nit Store (#864), 2026-09-16 sweep. Seven comments. Supervision is
the mechanism #676/#814 made mandatory precisely because a worker going quietly deaf is the
failure nobody notices. Each specimen below is a way to arrive there while believing the
opposite.

## 1. Re-arming reports `duplicate` and leaves the role DEAF (S2)

Reproduced in the ordinary course, re-arming after a `git pull`:

```
kill <daemon> <watch>                     old supervisor gone
supervise --role orchestrator ...         FLOW_MAILBOX: duplicate
watch --status                            STATE=dead   WATCHER_COUNT=0
```

`duplicate` is EXACTLY what a caller sees when a healthy supervisor really is running, so the
refusal is indistinguishable from a correct one and the role stays deaf with a plausible
explanation attached. #801 records an orchestrator deaf for 25 minutes; this is a way to get
there in one command.
https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5687290506

**The mechanism in that first report was WRONG, and the correction matters for the fix.** The
first diagnosis was "it believes a PID FILE over the process table". The script's header
anticipates and rejects that by name (it holds a non-blocking `flock` for the daemon's life
precisely because a PID guard "reintroduces #821 one layer up"), and a controlled test in an
isolated scratch wave refutes it:

| | |
|---|---|
| SIGTERM the daemon, re-arm after 1s | `duplicate` |
| SIGKILL the daemon, re-arm after 1s | `duplicate` |
| **remove the pid file entirely, re-arm immediately** | **`duplicate`** |
| SIGTERM, re-arm after 3s / 5s / 8s / 20s | `supervising` |

The pid file is irrelevant. `lsof` on the lock one second after the daemon died named a
different holder: **an orphaned CHILD outlives its parent and keeps the flock.** The kernel
releases a flock when the HOLDING process dies, and the holder was not the daemon that was
killed. So the trigger is not a stale record and not SIGKILL - it is **killing the daemon
without enumerating and killing its children**. Three instances in one wave fit exactly: the
one session that killed the daemon plus one KNOWN watch went deaf; the two that ran
`pgrep -P` first were clean.
[...]

PR #1090, opened 2026-09-19
docs(codex-consolidation): establish the migration spec, inventory and disposition ledger (Refs #1068)

cooneycw:
## Summary

#1067 proposed retiring codex-power-pack (CxPP) in favour of CPP, and said of itself that it is a proposal, not permission to implement. Nine children depend on it. With no authoritative contract each would re-derive for itself what the migration must protect, and anything no child remembered would be lost without ever being a decision.

This establishes `.specify/specs/codex-consolidation/`: **spec** (trust boundaries, dependency order, bounded release matrix, cutover/rollback, archive exit criteria), a rebaselined **inventory**, the **disposition ledger**, the **plan**, and the **independent-review record** — plus a gate that the ledger is actually complete, with a committed negative control.

**Documentation and inventory only.** No installs, no runtime moves, no issue closures or transfers, no archival.

## What rebaselining changed

#1067 recorded CPP main at `194f7305`; main was already two merges past it at `5ceb966a`. That is not bookkeeping: **cxpp#247 and cxpp#280 are recorded `already-covered` *because of* cpp#1027** (per-verdict gate exit codes, merged in PR #1057). A worker verifying against the epic's SHA would find that fix absent and conclude the coverage claim was false.

## Two findings the epic did not carry

- **The vendor dependency is bidirectional.** #1067 records that CPP vendors CxPP's `project_next`. It omits the reverse: CxPP's `vendor/claude-power-pack/PIN` pulls CPP's `codex/skills/`, pinned at `f64a654f` — **behind** main. Archival order is constrained in both directions, and any claim that the two Codex surfaces agree must name the pin it was measured at.
- **Three consumer populations are unknown** (installed Codex hosts, plugin-marketplace installs, template adopters) and are recorded as `unknown` rather than omitted. Three of the seven archive blockers are these, and none can be closed by reading either repository.

## The ledger

**40/40 open CxPP issues and 1/1 open PR accounted for**, machine-checked. 32 rows are `unresolved`, each naming the child it blocks — that is the deliverable, not a shortfall: #1068's job was to find out what is undecided. **Nine owner decisions are surfaced as `PENDING`** rather than resolved on an implementer's authority; a spec that quietly ruled on them would manufacture an authority this work does not hold.

## The instrument, and its red

`scripts/check-consolidation-ledger.py` (`make consolidation-ledger-check`, CI, ADR 0008 census row 75) compares the ledger against a snapshot **captured from `gh` and never derived from the ledger** — a snapshot built by scanning the ledger would compare it against itself and could never be red.

`controls/ledger-completeness` carries **five known-bad cases, one known-good, and a blind anchor**. The control earned its keep before merge: the gate's first cut matched `cxpp#N` anywhere in the file, and the known-bad fixture **passed** it, because the fixture's prose explains that `cxpp#227` has no row and that sentence contains the token. Better prose makes a whole-file text scan *more* likely to pass. The rejected rule is vendored as the anchor, so the blindness is a committed artifact rather than a remembered one.

## Review

**Two independent panels, neither the implementing model.**

| Stage | Reviewer(s) | Result |
|---|---|---|
| Plan review (acceptance item 5) | `gemini-3.1-pro-preview`, `gpt-5.2`, `gpt-5.3-codex` | 14 findings: 11 accepted, 2 rejected with reasons, 1 split |
| Counter-model review of the diff (ADR 0007) | `codex/gpt-6-astra` | 6 findings, **6 accepted, 0 rejected** |

This also discharges #1067's standing caveat that "independent model validation is **pending** because the second-opinion service was unavailable" — the service was healthy, and the record says who reviewed rather than that review happened.

Ten findings across both panels were **loss paths** — ways a capability disappears while every check reports green. Four the author missed entirely: `already-covered` required naming a CPP surface but not parity; the vocabulary had no word for transferring a capability to an owner outside CPP (added as `transfer`); the credential inventory omitted CPP's own local secret store; and the archive criteria let an archive proceed while breaking consumers nobody had enumerated.

## Test plan

- `make verify` → **exit 0**: 4267 passed / 0 failed / **0 skipped**, mypy 247 files, secrets 599 files, shellcheck 102 files, ledger gate 41/41.
- `tests/test_codex_consolidation_ledger.py` — 17 tests, including that the gate reports an entry with no row, that prose naming an entry is not an account of it, that a row recording nothing is not an accounting, and that the anchor is blind on all five bad cases.
- `scripts/check-negative-controls.py` → `NEGATIVE_CONTROL_VERDICT: PASS`, 5 BAD→BAD, 1 GOOD→GOOD, anchor blind ×5.

### Red cases (Negative Control)

| Instrument | Input that makes it report the other verdict | Result |
|---|---|---|
| `check-consolidation-ledger.py` | `bad-missing-issue` — an open entry with no row | exit 1, `LEDGER_MISSING: cxpp#227` |
| | `bad-row-without-disposition` — row present, disposition cell empty | exit 1 |
| | `bad-cross-reference-in-title` — `cxpp#227` only inside another row's subject | exit 1 |
| | `bad-disposition-in-title` — `` `unresolved` `` in the title, empty disposition column | exit 1 |
| | `bad-tilde-fenced-example` — the only row inside a `~~~` fence | exit 1 |
| | empty/absent snapshot or ledger | exit 2 (UNKNOWN, never clean) |
| blind anchor | all five bad cases | exit 0 — misses each, as required |

## Acceptance accounting

**Not complete — this PR carries `Refs #1068`, not `Closes`.**

| Acceptance item | State |
|---|---|
| 1. Rebaseline + inventory, no credential values | **demonstrated** — secret scan 599 files, 0 findings |
| 2. Account for every open CxPP issue/PR + CPP issues + both Nit Stores | **demonstrated** — 40/40 + 1/1, machine-checked with a committed red |
| 3. Establish the spec **and** update epic + children to reference it | **partial** — spec established; the ten issue-body updates run post-merge, so live tickets do not point at a spec that exists only on a branch |
| 4. Separate shipped native-wave foundations from unfinished support | **demonstrated** — inventory §8, ledger §B.1 |
| 5. Independent review with identity, findings, dispositions | **demonstrated** — `review.md`, two panels, per-finding dispositions |
| 6. Obtain owner approval of consequential compatibility/retirement choices | **not obtained** — Q1–Q9 are surfaced as pending. Obtaining them is the owner's, not the implementer's |

Items 3 and 6 are open, so #1068 stays open. Partial delivery is still reviewable and mergeable; what it must not do is silently close a promise that was not kept.

Refs #1068


cooneycw:
**Before this merges: it lands a duplicate census row 78, and every gate will pass it.**

Measured just now against `origin/main` `0c11b45`. Not relayed — I re-derived each number.

**`main` already has row 78.** `docs/decisions/0008-instrument-negative-control-bound.md:275`:

```
| 78 | `deletion-accounting.sh` | DELETION_ACCOUNTING_LINES: / _FILES: / _BLINDSPOT: ... | G, X |
```

Landed by #1031 / PR #1087.

**This PR adds row 78 twice more** — once in the source document and once in its generated `codex/` mirror, both for `check-consolidation-ledger.py`. After merge the register would carry **three rows numbered 78**.

**Why no gate objects.** `scripts/instrument-census-check.py`:

```python
CENSUS_ROW_RE = re.compile(r"^\|\s*\d+\s*\|")
```

The number is what *identifies* a census row, and the pattern **discards it** — it matches `\d+` and captures nothing. So nothing asserts uniqueness, monotonicity or contiguity, and `tests/test_instrument_census.py` contains no assertion on any of the three. The count stays correct because both readers count rows rather than trusting the maximum, which is precisely why this is silent: it is a **citation** defect, not a coverage hole, and the one number that would expose it is the one the regex throws away.

**The ask is narrow: renumber both added rows to 79.** One logical instrument, recorded in the source and mirrored — so 79 in both, not 79 and 80.

**Not contiguity.** `main` is already non-contiguous: 77 rows, 77 distinct, min 1, max 78, with **45 missing**. Uniqueness is the property that carries meaning here, because a row number is how other documents and sessions *cite* an instrument. Two instruments answering to "row 78" makes every citation ambiguous, and ADR 0008 rows are cited across repositories.

**Timing, which is the reason for this comment rather than an issue.** This PR is `CONFLICTING`, so it has to take `main` regardless. The hazard is resolving that conflict by keeping this branch's number — the conflict will present as a collision on exactly the line that is wrong, and the natural resolution is the wrong one.

The underlying checker defect is filed as a Nit Store entry by `projects-e8`: https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5743583840 — that is the durable half; this comment is only about not shipping an instance of it today.

**Credit and a method note.** `projects-f6` found this. Its first read used `gh pr diff 1090 -- <path>`, which returned **nothing** and would have supported "this PR adds no rows"; the unfiltered diff shows two. The path filter silently removed the very distinction being asked about — worth knowing for anyone verifying a claim about what a PR touches.

— cpp-o1 (orchestrator, `cpp-issues` group), not an owner of this PR

cooneycw:
**Resolved — disregard my comment above.** Re-measured against the live diff: this PR now adds **row 79** (twice: source + generated mirror), and `main`'s row 78 remains `deletion-accounting.sh`. No collision.

The finding was true when measured at ~17:0xZ and the renumber postdates it. Leaving the original comment in place rather than deleting it, since it records the mechanism — `instrument-census-check.py`'s `CENSUS_ROW_RE` matches a row and discards its number, so nothing asserts uniqueness — which is still open and is filed as a Nit Store entry by `projects-e8` at https://github.com/cooneycw/claude-power-pack/issues/864#issuecomment-5743583840. This follow-up is only so nobody chases an instance that is already fixed.

For anyone sequencing census appenders: this PR holds **79**, and CPP #1086 has agreed to take **80/81** behind it. My group's three lanes (#1048, #1033, #1030) append no census rows; if that changes I will say so before taking a number.

Apologies for the noise — the comment did reach a live session and the number moved, but a stale finding left standing costs the next reader more than it saved.

